PR #21, opened 2019-01-09
Encoding for aliased optionals

### LINKED ISSUE #19: Support for aliased optionals
There are currently a number of places we incorrectly handle alias types of optionals.

- Encoding methods: Today we would attempt to create encoding methods on whose receiver is a pointer type: this is not legal in go.
- Client responses: If an endpoint returns an optional, we incorrectly require the response to be non-nil.

To do this correctly, I think we need to generate a struct with a pointer field instead of an aliased pointer type. Example:
```go
// old
type OptionalDateAlias *datetime.DateTime
// new
type OptionalDateAlias struct {
	Value *datetime.DateTime
}
// unchanged
type ConcreteDateAlias datetime.DateTime
```

We will need to generate Marshal/Unmarshal methods which read/write from the value field. Clients will receive this container type and have to check `resp.Value != nil` instead of `resp != nil`. This has the added benefit that generated clients will return a concrete struct like they do with non-optional aliases, so we do not need a special case in the servicewriter.

Because this is more a result of the optional than the alias, I think we should leave non-optional aliases as-is for ease of use. The asymmetry should be palatable or even expected given the expectation that the response can be empty.

bmoylan:
Fixes #19 

This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/21)

### DIFF
--- a/conjure/aliaswriter.go
+++ b/conjure/aliaswriter.go
@@ -17,6 +17,7 @@ package conjure
 import (
 	"fmt"
 	"go/token"
+	"strings"
 
 	"github.com/palantir/goastwriter/astgen"
 	"github.com/palantir/goastwriter/decl"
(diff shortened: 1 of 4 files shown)


nmiyake (approved):
> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/21)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #33, opened 2019-01-22
Do not panic when processing Conjure with an external type


nmiyake:
When processing a valid Conjure definition that contains:

```
              "type" : "external",
              "external" : {
                "externalReference" : {
                  "name" : "ExampleExternalType",
                  "package" : "com.palantir.types"
                },
                "fallback" : {
                  "type" : "primitive",
                  "primitive" : "STRING"
                }
```

`conjure-go` panics with the following output:

```
panic: runtime error: index out of range

goroutine 1 [running]:
.../vendor/github.com/palantir/conjure-go/conjure/types.NewGoTypeFromExternalType(0xc0005d6140, 0x13, 0xc0005d6180, 0x19, 0xc0005347f0, 0x9, 0xc00057a9a0, 0x0, 0x0, 0x0, ...)
        .../vendor/github.com/palantir/conjure-go/conjure/types/types.go:226 +0x158
```

Because external types are on a path towards deprecation I don't think we need to add support for specifying external Go types, but at the very least we should try to parse and use the fallback type (and use the `any` type if a fallback doesn't exist).

If that's too much work, then a stopgap could be to just use the `any` type for now -- even that would be strictly better (while technically still fulfilling the spec) than panicking on a valid definition as we do now.

bwaldrep:
Fixed by #40

PR #40, opened 2019-03-20
Parse fallback when failing to parse external

bwaldrep:
Currently we panic if the type name in an external type does not include
a ":" (not required by the spec). Instead of panicing, we now try to parse
the fallback type and return that. Added test that repros the original
issue and confirms fallback to primitive types works.

Addresses #33

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/40)

### DIFF
--- a/conjure/types/types.go
+++ b/conjure/types/types.go
@@ -15,6 +15,7 @@
 package types
 
 import (
+	"errors"
 	"fmt"
 	"path"
 	"strings"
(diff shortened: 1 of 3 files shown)


bmoylan (approved):
Thanks! Looks good to me.  I reviewed the spec and I'm not sure where we got the colon-delimted thing from, possibly the internal version of conjure before the spec was rewritten for OSS.

Going to merge this to unblock you, but we probably need a more wholesale revisit of external types.

> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/40#-:-LaQjGH5BYimPyGbbxYa:b-tyqz59)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #41, opened 2019-04-02
Request body empty on client retry for clients using WithRawRequestBody


johnhferland:
See: https://github.com/palantir/conjure-go-runtime/issues/33

PR #42, opened 2019-04-02
Generate request body provider

johnhferland:
Use `WithRawRequestBodyProvider` instead of `WithRawRequestBody` so retries will have the body set.

Fixes: https://github.com/palantir/conjure-go/issues/41

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/42)

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -103,7 +103,7 @@
   revision = "10da8023aca66638ce70ae91d94e565b802f759d"
 
 [[projects]]
-  digest = "1:c16cb016436cfcc7d15d9e7a8a49bfe112656615719d53eafe0d0fcfef0facba"
+  digest = "1:954ea9a898c5e99d510dba4475bf43f79171f727de3bac42c6aff241598044e8"
   name = "github.com/palantir/conjure-go-runtime"
   packages = [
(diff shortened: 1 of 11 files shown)


johnhferland (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/42)* status: 0 of 11 files reviewed, 1 unresolved discussion

___
*[conjure/types/types.go, line 279 at r1](https://reviewable.io/reviews/palantir/conjure-go/42#-LbVEYY0BPfL5etIw2ap:-LbVEYY14Hp9oRqOc5yu:b-svo7dz) ([raw file](https://github.com/palantir/conjure-go/blob/f294ae305d72ce85baf3ff1ee36164c8c6477b5b/conjure/types/types.go#L279)):*
> ```Go
> }
> 
> type funcType struct {
> ```

I wanted to express `func() io.ReadCloser`, and the logic of goType didn't work for it so I defined a new function type. The complexity of supporting function types generically with inputs and outputs may be total overkill, but I couldn't think of a nice way to special case this.

johnhferland:
Closing this out for now while I work on fixing the integration tests

PR #45, opened 2019-04-23
Add test for malformed union

jdhenke:
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/45)


jdhenke:
Test will fail until we change what the generated code looks like and regenerate the code in `./integration_test/testgenerated` used by this test.

PR #70, opened 2019-07-11
Fix: add String() function for text alias types

### LINKED ISSUE #68: ResourceIdentifier aliases do not work as expected
## What happened?

An internal product defines a ResourceIdentifier type alias like so:

    // The RID for a Table.
    type TableRid rid.ResourceIdentifier

This `TableRid` type is used as a service query parameter:

    postEvent:
        http: POST /edits/tables/{tableRid}/events
        args:
            tableRid:
                docs: The identifier for the Table being edited.
                type: TableRid
                markers:
                    - Safe

which generates this interface function:

    PostEvent(ctx context.Context, authHeader bearertoken.Token, onBehalfOfArg *uuid.UUID, tableRidArg TableRid, tableEditedEventPostRequestArg TableEditedEventPostRequest) error

The client code generated uses `fmt.Sprint` on the `TableRid` object to path escape the parameter:

    requestParams = append(requestParams, httpclient.WithPathf("/storage/edits/tables/%s/events", url.PathEscape(fmt.Sprint(tableRidArg))))

`fmt.Sprint` does not call `TableRid.String()` because this does not exist on the alias. A generic string is used and the rid string ends up space delimited instead of what `ResourceIdentifier.String()` does.

Once the parameter is passed incorrectly the conjure-java based server returns errors because TableRid does not convert to a properly formatted rid string.

Aliased `ResourceIdentifier`s should carry forward the ResourceIdentifier.String() func.

nmiyake:
## Before this PR
`conjure-go` performs "PLAIN" serialization of types (used in scenarios like header and query parameters) by using `fmt.Sprint`. The types for `datetime` and `rid` implement `String()`, so this works fine.

However, types that are aliases of these types do not generate code for `String()`, so they do not use the custom string representation, and thus break.

## After this PR
==COMMIT_MSG==
Make it so that any alias type that defines a MarshalText function
also defines a String function that delegates to MarshalText.

Fixes #68
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/70)

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -170,7 +170,7 @@
   version = "v4.0.0"
 
 [[projects]]
-  digest = "1:ee5657d56e2c8a9f4cb15a219262585d8ff33bc778cd9e845dd008a34962d6cd"
+  digest = "1:d6a4e4b0c5bdcebdd8532af5916195853249ea10b0c6c5858fbc48960188d5a2"
   name = "github.com/palantir/pkg"
   packages = [
     "bearertoken",
@@ -194,8 +194,8 @@
     "uuid/internal/uuid",
   ]
   pruneopts = "UT"
-  revision = "7fd8d6e85a713f1ec5376c81b97c7ed1033c9816"
-  version = "0.10.0"
+  revision = "3c3292ada62a06084577c127e16e99b7a14675b5"
+  version = "0.10.5"
 
 [[projects]]
   digest = "1:89b7c2365c6b481a9e1c83a203c817d117376f48330811a977d84ec198281c8b"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -30,7 +30,7 @@
 
 [[constraint]]
   name = "github.com/palantir/pkg"
-  version = ">=0.10.0"
+  version = ">=0.10.5"
 
 [prune]
   go-tests = true
--- a/changelog/@unreleased/pr-70.v2.yml
+++ b/changelog/@unreleased/pr-70.v2.yml
@@ -1,0 +1,12 @@
+type: fix
+fix:
+  description: |-
+    Add String() function for text alias types
+
+    Make it so that any alias type that defines a MarshalText function
+    also defines a String function that delegates to the String() function
+    of the type being aliased.
+
+    Fixes #68
+  links:
+  - https://github.com/palantir/conjure-go/pull/70
--- a/conjure-go-verifier/conjure/verification/types/aliases.conjure.go
+++ b/conjure-go-verifier/conjure/verification/types/aliases.conjure.go
@@ -572,6 +572,10 @@ func (a *ReferenceAliasExample) UnmarshalYAML(unmarshal func(interface{}) error)
 type ListBinaryAliasExample [][]byte
 type UuidAliasExample uuid.UUID
 
+func (a UuidAliasExample) String() string {
+	return uuid.UUID(a).String()
+}
+
 func (a UuidAliasExample) MarshalText() ([]byte, error) {
 	return uuid.UUID(a).MarshalText()
 }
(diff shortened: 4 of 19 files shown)


nmiyake:
Alternative approaches considered:

* Define `String() string` in a manner that delegates to the underlying type. This generally works, but `bearertoken` does not currently implement `Stringer`, so we would need to implement it there to make it work in this case.
* Change the generated client code to use `MarshalText` rather than `fmt.Sprint` to perform "PLAIN" serialization
  * Makes sense (and may be "more correct"), but generated code is a bit of a pain -- either need to introspect on the type and generate proper `MarshalText()` vs. `fmt.Sprint` based on the type, or generate type assertion code that uses `MarshalText` only if the interface is implemented

Opted to go with the current approach (code-gen `String() string` with an implementation that delegates to `MarshalText()`) for now, but if you feel one of the other approaches above (or an alternate one altogether) is preferable, let me know.

nmiyake:
@bmoylan not urgent, but can you take a look when you get a chance?

bmoylan (commented):
> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/70)* status: all files reviewed, 1 unresolved discussion (waiting on @nmiyake)

___
*[conjure/aliaswriter.go, line 128 at r1](https://reviewable.io/reviews/palantir/conjure-go/70#-LjsA-6vAE9ulTkNNNk7:-LjsA-6vAE9ulTkNNNk8:b-ttfojh) ([raw file](https://github.com/palantir/conjure-go/blob/4924dbe914c58845d42c6861cbb9401ee01c977a/conjure/aliaswriter.go#L128)):*
> ```Go
> //
> //    func (a DateAlias) String() string {
> //        text, _ = a.MarshalText()
> ```

I'd be ok with this implementation, but if `bearertoken` is the only one that doesn't implement `fmt.Stringer` and we know the fixed set of types, I think I'd prefer we just fix that (https://github.com/palantir/pkg/pull/164) then generate using `.String()`. The ignored error is probably fine but makes me a little uneasy.

nmiyake (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/70)* status: all files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[conjure/aliaswriter.go, line 128 at r1](https://reviewable.io/reviews/palantir/conjure-go/70#-LjsA-6vAE9ulTkNNNk7:-LjvgU89CLJzBCshsFSW:b-5c1fc1) ([raw file](https://github.com/palantir/conjure-go/blob/4924dbe914c58845d42c6861cbb9401ee01c977a/conjure/aliaswriter.go#L128)):*

I'd be ok with this implementation, but if `bearertoken` is the only one that doesn't implement `fmt.Stringer` and we know the fixed set of types, I think I'd prefer we just fix that (https://github.com/palantir/pkg/pull/164) then generate using `.String()`. The ignored error is probably fine but makes me a little uneasy.
</blockquote></details>

I did consider that approach -- initially, I was a bit worried about requiring the downstream dependency bump, but I guess it's not that big of a deal. Based on this, here's my proposed plan:

* Add `String()` to `bearertoken` type in `pkg` and merge
* Add `String()` to `binary` type in `pkg` and merge
* Cut a release of `pkg`
* Update this PR to use implementation of `String()` that delegates to subtype

This should solve the "plain" serialization issue for almost all types -- the one exception is the infinities for `double` (the Conjure spec specifies that representation must be `+Infinity` or `-Infinity`, but our implementation currently produces `+Inf` or `-Inf`). To fully solve the issue, we should probably just code-gen the code that performs "plain" serialization for these types. That's a bit more involved, so will file a separate issue to track that.

nmiyake (commented):
Ready for another review

> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/70#-:-Ljw4J26DHvfaZww1l1D:b-a0pjw2)* status: 3 of 19 files reviewed, 1 unresolved discussion (waiting on @bmoylan)

bmoylan (approved):
:lgtm: thanks!

> Reviewed 16 of 16 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/70#-:-Ljw6Mcq-CyF0zipr9nk:brfqlux)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #74, opened 2019-07-16
Change release version to 4.0.3

nmiyake:
## After this PR
==COMMIT_MSG==

==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/74)


nmiyake:
I performed a release of this repository using `auto-release`, but didn't realize that it doesn't support `v` prefixes until after the release was performed :(

Plan of action:

* Merge this PR that moves changelogs to proper location
* Manually create a 4.0.3 release
* Update release notes for 2.0.1 to note that the release was performed by mistake and this version should not be used
* File issue on auto-release to support configuring version prefix

bmoylan:
Alternatively, should we just stop using the v prefix? Not sure it gets us anything, especially since the infrastructure now doesn't support it nicely.

👍

bmoylan (approved):

nmiyake:
The main reason we use `v` prefixing is in anticipation of this eventually switching to be a module -- Go modules only recognizes tags that start with a `v` as semantic versions (see vgo spec and https://github.com/golang/go/issues/24007#issuecomment-377603850).

bmoylan:
Ah ok, does that mean anything wanting to use go modules can not use autorelease? that would be sad

nmiyake:
I've filed an issue on `auto-release` to support this use case (either via config or some other mechanism). This shouldn't be too hard to support from a technical perspective given that tag prefixes are already supported for things like frontend monorepos -- in that scenario, the project name is used as a tag prefix (for example, `@foo-project@1.2.3`), and there's code to find and sort tags of that form. Conceptually, treating `v` as such a prefix is pretty much the same thing.

bmoylan:
cool, thanks for explaining!

PR #76, opened 2019-07-17
Fix: Construct client with TokenProvider

### LINKED ISSUE #75: FR: Construct client with TokenProvider
We currently generate client interfaces with and without auth token arguments. In cases where we use short-lived tokens, we currently must provide a token to each request. 

Instead, we should generate a second implementation for the `*WithAuth` interface which can be constructed with a [`httpclient.TokenProvider`](https://godoc.org/github.com/palantir/conjure-go-runtime/conjure-go-client/httpclient#TokenProvider).

Blue9:
## Before this PR

The generated client interface needs a new auth token every time we make a request. The *WithAuth interface takes an auth token but this falls short when the token is short-lived. Having a client that stores its own token provider would make it easier to make requests.

## After this PR
==COMMIT_MSG==
Add client with token provider

Fixes: #75 
==COMMIT_MSG==

A *WithTokenProvider interface will be generated in the case that all endpoints that require authentication are of the same auth type (header or cookie) (See `canAddTokenInterface` in `servicewriter.go`).

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/76)


palantirtech:
Thanks for your interest in [palantir/conjure-go](https://github.com/palantir/conjure-go), @Blue9! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

conjure/servicewriter.go
```diff
@@ -43,14 +43,23 @@ const (
 	httpClientImportPath = "github.com/palantir/conjure-go-runtime/conjure-go-client/httpclient"
 	httpClientClientType = expression.Type("httpclient.Client")
 	httpClientPkgName    = "httpclient"
+
+	tokenProviderVar        = "tokenProvider"
+	tokenProviderType       = "httpclient.TokenProvider"
```
nmiyake: Value should be wrapped in `expression.Type()` (analogous to `httpClientClientType`)

conjure/servicewriter.go
```diff
@@ -43,14 +43,23 @@ const (
...
+	tokenProviderImportPath = "github.com/palantir/conjure-go-runtime/conjure-go-client/httpclient"
 )
 
+type config struct {
```
nmiyake: Type name should be something more specific like `astServiceConfig` -- the `conjure` package is fairly broad, so the type name should be more specific than just `config`.

conjure/servicewriter.go
```diff
@@ -43,14 +43,23 @@ const (
...
+	withTokenProvider bool
+}
+
 func astForService(serviceDefinition spec.ServiceDefinition, info types.PkgInfo) ([]astgen.ASTDecl, StringSet, error) {
 	allImports := NewStringSet()
 	serviceName := serviceDefinition.ServiceName.Name
 	isClient := true
-
-	interfaceAST, imports, err := serviceInterfaceAST(serviceDefinition, info, false, isClient)
+	interfaceAST, imports, err := serviceInterfaceAST(serviceDefinition, info, config{withAuth: false, isClient: isClient})
```
nmiyake: Why does this inline struct initialization set `withAuth: false` but not set `withTokenProvider`? The initialization style should be consistent: either skip initializing fields that will have the default value (in which case the `withAuth` initialization should just be removed), or explicitly initialize every field. My preference would be the former.

conjure/servicewriter.go
```diff
@@ -134,13 +168,41 @@ func hasAuth(endpoints []spec.EndpointDefinition) (hasHeaderAuth, hasCookieAuth
...
+			return false
+		}
+		if possibleHeaderAuth != nil {
+			numHeader++
+		}
+		if possibleCookieAuth != nil {
+			numCookie++
+		}
+		if numHeader != 0 && numCookie != 0 {
+			return false
```
nmiyake, line 178: Why is it the case that a token interface can't be added if a service has endpoints that perform both cookie and header auth?

bmoylan (changes requested):
> Reviewed 6 of 8 files at r7, 1 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/76)* status: 7 of 8 files reviewed, 3 unresolved discussions (waiting on @Blue9)

___
*[conjure/servicewriter.go, line 52 at r8](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk-bZR3Dr2Uk8UWCp7Q:-Lk-bZR3Dr2Uk8UWCp7R:bp79pj4) ([raw file](https://github.com/palantir/conjure-go/blob/e3395f213a1b4561a12403c22c79bac9742b0f5d/conjure/servicewriter.go#L52)):*
> ```Go
> )
> 
> type config struct {
> ```

Please use a more descriptive struct name, maybe `serviceASTConfig` or something
___
*[conjure/servicewriter.go, line 122 at r8](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk-eyRx4A-Py5RDewO7:-Lk-eyRx4A-Py5RDewO8:b-2gy139) ([raw file](https://github.com/palantir/conjure-go/blob/e3395f213a1b4561a12403c22c79bac9742b0f5d/conjure/servicewriter.go#L122)):*
> ```Go
> 
> 	if canAddTokenInterface(serviceDefinition.Endpoints) {
> 		withTokenInterfaceAST, imports, err := serviceInterfaceAST(serviceDefinition, info, config{isClient: isClient, withTokenProvider: true})
> ```

Why do we need to regenerate this interface instead of just providing a second implementation for the existing `WithAuth` interface? Are there cases where they'll be different?
___
*[conjure/servicewriter.go, line 171 at r8](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk-eFw0EPEyL3nd3wSO:-Lk-eFw0EPEyL3nd3wSP:bo3ofxx) ([raw file](https://github.com/palantir/conjure-go/blob/e3395f213a1b4561a12403c22c79bac9742b0f5d/conjure/servicewriter.go#L171)):*
> ```Go
> }
> 
> func canAddTokenInterface(endpoints []spec.EndpointDefinition) bool {
> ```

Can you add a comment explaining what this is checking?

conjure/servicewriter.go
```diff
@@ -134,13 +168,41 @@ func hasAuth(endpoints []spec.EndpointDefinition) (hasHeaderAuth, hasCookieAuth
 	return
 }
 
-func serviceInterfaceAST(serviceDefinition spec.ServiceDefinition, info types.PkgInfo, withAuth, isClient bool) (astgen.ASTDecl, StringSet, error) {
+func canAddTokenInterface(endpoints []spec.EndpointDefinition) bool {
```
nmiyake, line 161: Add a comment that semantically describes what this function does

nmiyake (changes requested):

conjure/servicewriter.go
Blue9: There's no reason it couldn't but it's not clear to me how one token provider would be useful for generating both types of tokens. That being said, let me know if we want to change this to add a token interface on the same condition as the WithAuth interface.

Blue9 (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/76)* status: 4 of 8 files reviewed, 8 unresolved discussions (waiting on @bmoylan and @nmiyake)

___
*[conjure/servicewriter.go, line 48 at r8](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk-NU_SkjoX4FsH90UQ-r8-48:-Lk0N0ef4iac1ONnoPPy:b-896fix) ([raw file](https://github.com/palantir/conjure-go/blob/e3395f213a1b4561a12403c22c79bac9742b0f5d/conjure/servicewriter.go#L48)):*

Value should be wrapped in `expression.Type()` (analogous to `httpClientClientType`)
</blockquote></details>

Done.
___
*[conjure/servicewriter.go, line 52 at r8](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk-NU_SkjoX4FsH90UQ-r8-52:-Lk0N21G2azICB6EgbTf:b-896fix) ([raw file](https://github.com/palantir/conjure-go/blob/e3395f213a1b4561a12403c22c79bac9742b0f5d/conjure/servicewriter.go#L52)):*

Type name should be something more specific like `astServiceConfig` -- the `conjure` package is fairly broad, so the type name should be more specific than just `config`.
</blockquote></details>

Done.
___
*[conjure/servicewriter.go, line 52 at r8](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk-bZR3Dr2Uk8UWCp7Q:-Lk0N2Tg1EGktSYXFzop:b-896fix) ([raw file](https://github.com/palantir/conjure-go/blob/e3395f213a1b4561a12403c22c79bac9742b0f5d/conjure/servicewriter.go#L52)):*

Please use a more descriptive struct name, maybe `serviceASTConfig` or something
</blockquote></details>

Done.
___
*[conjure/servicewriter.go, line 62 at r8](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk-NU_SkjoX4FsH90UQ-r8-62:-Lk0N3KW5ULHjfoO8o7S:b-kbmv7q) ([raw file](https://github.com/palantir/conjure-go/blob/e3395f213a1b4561a12403c22c79bac9742b0f5d/conjure/servicewriter.go#L62)):*

Why does this inline struct initialization set `withAuth: false` but not set `withTokenProvider`? The initialization style should be consistent: either skip initializing fields that will have the default value (in which case the `withAuth` initialization should just be removed), or explicitly initialize every field. My preference would be the former.
</blockquote></details>

Agreed, going with former.
___
*[conjure/servicewriter.go, line 122 at r8](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk-eyRx4A-Py5RDewO7:-Lk0N6HK1a2AcopHrGfp:b-przqrs) ([raw file](https://github.com/palantir/conjure-go/blob/e3395f213a1b4561a12403c22c79bac9742b0f5d/conjure/servicewriter.go#L122)):*

Why do we need to regenerate this interface instead of just providing a second implementation for the existing `WithAuth` interface? Are there cases where they'll be different?
</blockquote></details>

Removed token provider interface.
___
*[conjure/servicewriter.go, line 171 at r8](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk-NU_SkjoX4FsH90UQ-r8-171:-Lk0N9Cb1Wf76ltCcmdu:b-896fix) ([raw file](https://github.com/palantir/conjure-go/blob/e3395f213a1b4561a12403c22c79bac9742b0f5d/conjure/servicewriter.go#L171)):*

Add a comment that semantically describes what this function does
</blockquote></details>

Done.
___
*[conjure/servicewriter.go, line 171 at r8](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk-eFw0EPEyL3nd3wSO:-Lk0N9X24bdTOVNCh_TZ:b-896fix) ([raw file](https://github.com/palantir/conjure-go/blob/e3395f213a1b4561a12403c22c79bac9742b0f5d/conjure/servicewriter.go#L171)):*

Can you add a comment explaining what this is checking?
</blockquote></details>

Done.

conjure/servicewriter.go
```diff
@@ -134,13 +161,44 @@ func hasAuth(endpoints []spec.EndpointDefinition) (hasHeaderAuth, hasCookieAuth
...
+// one endpoint has auth. The same auth type is required because a single token provider will likely not be useful for
+// both auth types.
+func canAddTokenInterface(endpoints []spec.EndpointDefinition) bool {
+	numHeader := 0
+	numCookie := 0
+	for _, endpointDefinition := range endpoints {
+		if endpointDefinition.Auth == nil {
+			continue
+		}
+		possibleHeaderAuth, err := visitors.GetPossibleHeaderAuth(*endpointDefinition.Auth)
```
nmiyake, line 167: Keep related logic together for readability. Example:

```go
if possibleHeaderAuth, err := visitors.GetPossibleHeaderAuth(*endpointDefinition.Auth); err != nil {
   return false
} else if possibleHeaderAuth {
    hasHeaderAuthEndpoint = true
}
```

conjure/servicewriter.go
nmiyake: OK, makes sense.

nmiyake (changes requested):

conjure/servicewriter.go
Blue9: 👍

bmoylan:
___
*[changelog/@unreleased/pr-76.v2.yml, line 1 at r11](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk1ClhXCoo-_hB-3dzd:-Lk1ClhXCoo-_hB-3dze:b8a2yct) ([raw file](https://github.com/palantir/conjure-go/blob/ba3ab153bc3712ed70bbb89aa43207a7ba836480/changelog/@unreleased/pr-76.v2.yml#L1)):*
> ```YAML
> type: fix
> ```

I'd probably call this a `feature`

Blue9 (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/76)* status: 3 of 8 files reviewed, 9 unresolved discussions (waiting on @bmoylan and @nmiyake)

___
*[changelog/@unreleased/pr-76.v2.yml, line 1 at r11](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk1ClhXCoo-_hB-3dzd:-Lk4VQdiBXiUnstma3C2:b-896fix) ([raw file](https://github.com/palantir/conjure-go/blob/ba3ab153bc3712ed70bbb89aa43207a7ba836480/changelog/@unreleased/pr-76.v2.yml#L1)):*

I'd probably call this a `feature`
</blockquote></details>

Done.

nmiyake (changes requested):
> Reviewed 3 of 4 files at r9, 2 of 2 files at r11, 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/76)* status: all files reviewed, 9 unresolved discussions (waiting on @Blue9 and @bmoylan)

___
*[changelog/@unreleased/pr-76.v2.yml, line 2 at r12](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk5x2Ip591CR1Y3OFRD:-Lk5x2Ip591CR1Y3OFRE:bu8nf7l) ([raw file](https://github.com/palantir/conjure-go/blob/57e8aa3ae8835527c442ddf2212d8f65dfe899d8/changelog/@unreleased/pr-76.v2.yml#L2)):*
> ```YAML
> type: feature
> fix:
> ```

This field also needs to be renamed to be `feature` (otherwise, this is not a valid union type)
___
*[conjure/servicewriter.go, line 169 at r9](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk5xeij8U6ZSQ3mAM2v:-Lk5xeij8U6ZSQ3mAM2w:blovwxc) ([raw file](https://github.com/palantir/conjure-go/blob/ba0898abce182a3a329576e2fa8541c1db603374/conjure/servicewriter.go#L169)):*
> ```Go
> func canAddTokenInterface(endpoints []spec.EndpointDefinition) bool {
> 	numHeader := 0
> 	numCookie := 0
> ```

These values are never used as numbers -- although they are incremented, you only ever check whether they are 0 or non-0. Semantically, it makes more sense for these to be booleans:

```go
var hasHeaderAuth, hasCookieAuth bool
```
___
*[conjure/servicewriter.go, line 182 at r9](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk5xu1M8xSc3T9iqRDn:-Lk5xu1M8xSc3T9iqRDo:bzbey6r) ([raw file](https://github.com/palantir/conjure-go/blob/ba0898abce182a3a329576e2fa8541c1db603374/conjure/servicewriter.go#L182)):*
> ```Go
> 			return false
> 		}
> 		if possibleHeaderAuth != nil {
> ```

Once you switch to using `bool`, this can change to:

```
hasHeaderAuth = hasHeaderAuth || possibleHeaderAuth != nil
```
___
*[conjure/servicewriter.go, line 188 at r9](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk5y5Jx1I78tDLrF8fR:-Lk5y5Jx1I78tDLrF8fS:bhynlrx) ([raw file](https://github.com/palantir/conjure-go/blob/ba0898abce182a3a329576e2fa8541c1db603374/conjure/servicewriter.go#L188)):*
> ```Go
> 			numCookie++
> 		}
> 		if numHeader != 0 && numCookie != 0 {
> ```

Then this can change to `if hasHeaderAuth || hasCookieAuth`
___
*[conjure/servicewriter.go, line 192 at r9](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk5yASY88eafLE7yWgK:-Lk5yASY88eafLE7yWgL:bl841b4) ([raw file](https://github.com/palantir/conjure-go/blob/ba0898abce182a3a329576e2fa8541c1db603374/conjure/servicewriter.go#L192)):*
> ```Go
> 		}
> 	}
> 	return numHeader > 0 || numCookie > 0
> ```

And this can be `return hasHeaderAuth || hasCookieAuth`

nmiyake (commented):
Sorry for being slow with the last review -- think this is pretty close; just a few more points of feedback.

> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/76#-:-Lk5yX1XE47Mnb18KS9j:b-5b261s)* status: all files reviewed, 9 unresolved discussions (waiting on @Blue9 and @bmoylan)

Blue9 (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/76)* status: 6 of 8 files reviewed, 9 unresolved discussions (waiting on @bmoylan and @nmiyake)

___
*[changelog/@unreleased/pr-76.v2.yml, line 2 at r12](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk5x2Ip591CR1Y3OFRD:-Lk6-REX6IC6Bmk6jkng:b-896fix) ([raw file](https://github.com/palantir/conjure-go/blob/57e8aa3ae8835527c442ddf2212d8f65dfe899d8/changelog/@unreleased/pr-76.v2.yml#L2)):*

This field also needs to be renamed to be `feature` (otherwise, this is not a valid union type)
</blockquote></details>

Done.
___
*[conjure/servicewriter.go, line 169 at r9](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk5xeij8U6ZSQ3mAM2v:-Lk6-S6S2jQ47nSG8V-Z:b-896fix) ([raw file](https://github.com/palantir/conjure-go/blob/ba0898abce182a3a329576e2fa8541c1db603374/conjure/servicewriter.go#L169)):*

These values are never used as numbers -- although they are incremented, you only ever check whether they are 0 or non-0. Semantically, it makes more sense for these to be booleans:

```go
var hasHeaderAuth, hasCookieAuth bool
```
</blockquote></details>

Done.
___
*[conjure/servicewriter.go, line 182 at r9](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk5xu1M8xSc3T9iqRDn:-Lk6-aJB02paBPdXhZXV:b-clzqy1) ([raw file](https://github.com/palantir/conjure-go/blob/ba0898abce182a3a329576e2fa8541c1db603374/conjure/servicewriter.go#L182)):*

Once you switch to using `bool`, this can change to:

```
hasHeaderAuth = hasHeaderAuth || possibleHeaderAuth != nil
```
</blockquote></details>

Done. Also moved the declaration to before the if-statement to pass `godel check`.
___
*[conjure/servicewriter.go, line 188 at r9](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk5y5Jx1I78tDLrF8fR:-Lk6-cvs0B84u6HmUTC4:b-vciuru) ([raw file](https://github.com/palantir/conjure-go/blob/ba0898abce182a3a329576e2fa8541c1db603374/conjure/servicewriter.go#L188)):*

Then this can change to `if hasHeaderAuth || hasCookieAuth`
</blockquote></details>

Done. (Changed to `hasHeaderAuth && hasCookieAuth`)
___
*[conjure/servicewriter.go, line 192 at r9](https://reviewable.io/reviews/palantir/conjure-go/76#-Lk5yASY88eafLE7yWgK:-Lk6-mrB3veENZ-SrdB6:b-896fix) ([raw file](https://github.com/palantir/conjure-go/blob/ba0898abce182a3a329576e2fa8541c1db603374/conjure/servicewriter.go#L192)):*

And this can be `return hasHeaderAuth || hasCookieAuth`
</blockquote></details>

Done.

nmiyake (approved):
> Reviewed 3 of 8 files at r7, 2 of 2 files at r13.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/76)* status: all files reviewed, 4 unresolved discussions (waiting on @bmoylan)

nmiyake:
@bmoylan I believe your feedback has been addressed -- can you take a look and confirm when you get a chance?

If I don't see an update in 24 hours, then will merge and any feedback/concerns can be addressed post-merge.

andybradshaw:
@bmoylan @nmiyake bump on this... don't want it sit around for too long

nmiyake:
Sorry, forgot to follow up after 24 hours -- merging.

bmoylan:
Thanks Nick! Sorry I didn't see this (just got back from two weeks ooto)

PR #77, opened 2019-07-17
Use named type for generating server or client

nmiyake:
Separating the functions makes more semantic sense given that, as
currently used, all callers use a hard-coded constant when calling
the function. If the caller always knows the "mode" of the function
being called at compile-time, it's good signal that these should just
be separate functions (which also reduces number of parameters and
makes the behavior clearer based on the name).

Only refactors internal private functions.

==NO_CHANGELOG==

## After this PR
==COMMIT_MSG==

==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/77)


bmoylan (approved):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/77)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @k-simons)

PR #99, opened 2019-11-24
Fix: ensure import paths are correct in modules

### LINKED ISSUE #98: Conjure generation produces incorrect imports if output is in Go module
## What happened?
When running conjure generation on a definition that produces services and structs in a project that is set up as a Go module, the imports in the generated Go code used import paths that started with `_/Volumes/...`, which are invalid.

The import paths should be correct (should be the module path + relative path from module root to the package)

nmiyake:
If the generated code is within a module, ensure that the import
paths are relative to the module root.

Fixes #98

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/99)


nmiyake (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/99)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @bmoylan and @TDAbboud)

___
*[conjure/conjure.go, line 367 at r2](https://reviewable.io/reviews/palantir/conjure-go/99#-LuQvxFm-o4UEVXs3NZ-:-LuQvxFm-o4UEVXs3NZ0:bomm7tm) ([raw file](https://github.com/palantir/conjure-go/blob/af7efedfb1d58b4dca9aa0b6dc84acd4c71cec50/conjure/conjure.go#L367)):*
> ```Go
> 	output, err := cmd.CombinedOutput()
> 	if err != nil {
> 		if string(output) == "go list -m: not using modules\n" {
> ```

Need to make a trade-off here, and open to feedback on it.

The risk of the current implementation is that it treats the output message as API, which is not guaranteed to be stable in the long-term.

The other approach would be to treat any error in execution as "not using modules" and simply proceed down the package logic at that point. Downside there is that if the command fails for some unexpected reason ("go" command does not exist, lack of permissions, something else), it is treated the same as the expected failure and silently goes down the other execution path.

Ended up going with this approach because the failure is more visible to the user -- if the message changes in the future, users generating in module mode will get a concrete failure (that includes the error message) and we can quickly move to fix it, while in the other mode the failure will be silent and go down another execution path without providing clear feedback.

bmoylan (commented):
> Reviewed 1 of 3 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/99)* status: all files reviewed, 3 unresolved discussions (waiting on @nmiyake and @TDAbboud)

___
*[conjure/conjure.go, line 367 at r2](https://reviewable.io/reviews/palantir/conjure-go/99#-LuQvxFm-o4UEVXs3NZ-:-LuY-dZT-mztCOW9lCJ7:b-p6uh2k) ([raw file](https://github.com/palantir/conjure-go/blob/af7efedfb1d58b4dca9aa0b6dc84acd4c71cec50/conjure/conjure.go#L367)):*

Need to make a trade-off here, and open to feedback on it.

The risk of the current implementation is that it treats the output message as API, which is not guaranteed to be stable in the long-term.

The other approach would be to treat any error in execution as "not using modules" and simply proceed down the package logic at that point. Downside there is that if the command fails for some unexpected reason ("go" command does not exist, lack of permissions, something else), it is treated the same as the expected failure and silently goes down the other execution path.

Ended up going with this approach because the failure is more visible to the user -- if the message changes in the future, users generating in module mode will get a concrete failure (that includes the error message) and we can quickly move to fix it, while in the other mode the failure will be silent and go down another execution path without providing clear feedback.
</blockquote></details>

If the API changes it will be in a new go release (hopefully minor version and never patch) so I think we are safe as long as consumers are not using newer go than this library has been tested/built with. As long as we stay up to date on circle configs I think we will catch most breaks.

For the "go not found" case, we could check for an `exec.Error` with `exec.ErrNotFound` in the `Err` field which is populated by LookPath before the command starts.
___
*[conjure/conjure.go, line 344 at r3](https://reviewable.io/reviews/palantir/conjure-go/99#-LuY1Xd94vfxk1b_la0y:-LuY1XdA4ICx2T7eikL2:b-d1df98) ([raw file](https://github.com/palantir/conjure-go/blob/cbabc43ca9d3e02ae3a7d801b2f4646277b92046/conjure/conjure.go#L344)):*
> ```Go
> 			return "", errors.Wrapf(err, "failed to determine relative path for module directory")
> 		}
> 		return path.Join(modName, relPath), nil
> ```

`filepath.Join`
___
*[conjure/conjure.go, line 381 at r3](https://reviewable.io/reviews/palantir/conjure-go/99#-LuY2-HQ0wVRr8zfkIZy:-LuY2-HQ0wVRr8zfkIZz:b-bcu909) ([raw file](https://github.com/palantir/conjure-go/blob/cbabc43ca9d3e02ae3a7d801b2f4646277b92046/conjure/conjure.go#L381)):*
> ```Go
> 	}{}
> 	if err := json.Unmarshal(output, &modJSON); err != nil {
> 		return "", "", err
> ```

maybe wrap this error with "failed to unmarshal go list -m output as json" or something to indicate what it was trying to do

nmiyake (commented):
Thanks for the review! Addressed PR feedback.

> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/99#-:-LuYC_Ce8OWImjFaV-D4:b-u7mom1)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @bmoylan and @TDAbboud)

___
*[conjure/conjure.go, line 367 at r2](https://reviewable.io/reviews/palantir/conjure-go/99#-LuQvxFm-o4UEVXs3NZ-:-LuYC8yhESFvkQckCaEa:b-5b3u1u) ([raw file](https://github.com/palantir/conjure-go/blob/af7efedfb1d58b4dca9aa0b6dc84acd4c71cec50/conjure/conjure.go#L367)):*

If the API changes it will be in a new go release (hopefully minor version and never patch) so I think we are safe as long as consumers are not using newer go than this library has been tested/built with. As long as we stay up to date on circle configs I think we will catch most breaks.

For the "go not found" case, we could check for an `exec.Error` with `exec.ErrNotFound` in the `Err` field which is populated by LookPath before the command starts.
</blockquote></details>

Yeah I think this approach will likely be good for now. Thanks for the extra info, though -- good to know how to handle that case if we decide to do that in the future.
___
*[conjure/conjure.go, line 344 at r3](https://reviewable.io/reviews/palantir/conjure-go/99#-LuY1Xd94vfxk1b_la0y:-LuYC6tv3E6rd8n3dhnC:b-896fix) ([raw file](https://github.com/palantir/conjure-go/blob/cbabc43ca9d3e02ae3a7d801b2f4646277b92046/conjure/conjure.go#L344)):*

`filepath.Join`
</blockquote></details>

Done.
___
*[conjure/conjure.go, line 381 at r3](https://reviewable.io/reviews/palantir/conjure-go/99#-LuY2-HQ0wVRr8zfkIZy:-LuYCFvXB7IHxwQCRZxK:b-896fix) ([raw file](https://github.com/palantir/conjure-go/blob/cbabc43ca9d3e02ae3a7d801b2f4646277b92046/conjure/conjure.go#L381)):*

maybe wrap this error with "failed to unmarshal go list -m output as json" or something to indicate what it was trying to do
</blockquote></details>

Done.

bmoylan (approved):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/99)* status: all files reviewed, 1 unresolved discussion (waiting on @nmiyake and @TDAbboud)

___
*[conjure/conjure.go, line 367 at r2](https://reviewable.io/reviews/palantir/conjure-go/99#-LuQvxFm-o4UEVXs3NZ-:-LuYEMQg4POH5r3FRuS8:b-8fx2u7) ([raw file](https://github.com/palantir/conjure-go/blob/af7efedfb1d58b4dca9aa0b6dc84acd4c71cec50/conjure/conjure.go#L367)):*

Yeah I think this approach will likely be good for now. Thanks for the extra info, though -- good to know how to handle that case if we decide to do that in the future.
</blockquote></details>

:+1:

PR #102, opened 2019-11-27
Fix: update godel-conjure-plugin and godel/pkg/products modules

nmiyake:
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/102)


nmiyake:
Minor dependency-only change that I'd like to push through to test some module dependency things -- merging.

PR #105, opened 2020-01-28
String lists and sets as query parameters

k-simons:
## Before this PR

- Query params of any list/set type would throw an error for sever generation

## After this PR
- Query params of the string list/set type will generate working server code
==COMMIT_MSG==

==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/105)


k-simons:
@nmiyake for Review

nmiyake (changes requested):
Thanks! Content generally looks good, although I have minor feedback on the codegen.

Can you add support for the other primitives as well as part of this PR? If not, let me know and I can modify the PR to add this. Since we're taking the time to add support for string collections now, I feel like we might as well just make the extra effort to support all primitive collections (doing so as part of this will likely be more efficient than waiting until somebody needs support for a non-list collection)

> Reviewed 5 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/105#-:-Lzh9vyZ7HdWhEqaar9m:b-3wujeh)* status: 5 of 7 files reviewed, 3 unresolved discussions (waiting on @ferozco and @k-simons)

___
*[changelog/@unreleased/pr-105.v2.yml, line 3 at r2](https://reviewable.io/reviews/palantir/conjure-go/105#-Lzh9dUNDjjaT8AjOvMr:-Lzh9dUNDjjaT8AjOvMs:b-1yd9ie) ([raw file](https://github.com/palantir/conjure-go/blob/14cd38971769cb4e2d83bb1819f2b9f402d8d953/changelog/@unreleased/pr-105.v2.yml#L3)):*
> ```YAML
> type: fix
> fix:
>   description: String lists and sets as query parameters
> ```

Can you make this a bit more descriptive? Since this is read by people in the release notes, would be nice to make this a full sentence (even if it's minor like "Add support for list<string> and set<string> query parameters.")
___
*[conjure/serverwriter.go, line 701 at r2](https://reviewable.io/reviews/palantir/conjure-go/105#-Lzh7zR18qTDGOccRRl8:-Lzh7zR22ZUL2TQDSdGk:bstj6k9) ([raw file](https://github.com/palantir/conjure-go/blob/14cd38971769cb4e2d83bb1819f2b9f402d8d953/conjure/serverwriter.go#L701)):*
> ```Go
> 		ifErrNotNilReturnErrStatement("err", nil)
> 		// type-specific unmarshal behavior
> 		// TODO(bmoylan): non string lists are unimplemented right now, but we _could_ iterate through the raw map and pull them out.
> ```

Would it be much more effort to just implement this properly? Since we're already pulling out the list, I think you could iterate over the elements and then convert them again, right?
___
*[conjure/visitors/visit_string_param.go, line 282 at r2](https://reviewable.io/reviews/palantir/conjure-go/105#-Lzh8NS12r37P62yfJYp:-Lzh8NS12r37P62yfJYq:b-saqds1) ([raw file](https://github.com/palantir/conjure-go/blob/14cd38971769cb4e2d83bb1819f2b9f402d8d953/conjure/visitors/visit_string_param.go#L282)):*
> ```Go
> 		LHS: []astgen.ASTExpr{
> 			expression.VariableVal(v.argName),
> 			expression.VariableVal("_"),
> ```

Should just omit this since we don't use the value and it's not necessary (can do a map lookup with only one value on LHS if we're not using the second bool)

k-simons (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/105)* status: 5 of 7 files reviewed, 3 unresolved discussions (waiting on @nmiyake)

___
*[changelog/@unreleased/pr-105.v2.yml, line 3 at r2](https://reviewable.io/reviews/palantir/conjure-go/105#-Lzh9dUNDjjaT8AjOvMr:-LzhIBrp7k5A1o985X24:bb6un1j) ([raw file](https://github.com/palantir/conjure-go/blob/14cd38971769cb4e2d83bb1819f2b9f402d8d953/changelog/@unreleased/pr-105.v2.yml#L3)):*

Can you make this a bit more descriptive? Since this is read by people in the release notes, would be nice to make this a full sentence (even if it's minor like "Add support for list<string> and set<string> query parameters.")
</blockquote></details>

Done
___
*[conjure/serverwriter.go, line 701 at r2](https://reviewable.io/reviews/palantir/conjure-go/105#-Lzh7zR18qTDGOccRRl8:-LzhIF2C7KMnAA6uLayv:b-pcu8a1) ([raw file](https://github.com/palantir/conjure-go/blob/14cd38971769cb4e2d83bb1819f2b9f402d8d953/conjure/serverwriter.go#L701)):*

Would it be much more effort to just implement this properly? Since we're already pulling out the list, I think you could iterate over the elements and then convert them again, right?
</blockquote></details>

So at a high level, I think that you would want to refactor `ParseStringParam` into just being `ParseParam` and then correctly use the internal visitor that it has. 

I poked at it and it was more effort than I was willing to give
___
*[conjure/visitors/visit_string_param.go, line 282 at r2](https://reviewable.io/reviews/palantir/conjure-go/105#-Lzh8NS12r37P62yfJYp:-LzhIXDm-ajcjot6sUHM:b-896fix) ([raw file](https://github.com/palantir/conjure-go/blob/14cd38971769cb4e2d83bb1819f2b9f402d8d953/conjure/visitors/visit_string_param.go#L282)):*

Should just omit this since we don't use the value and it's not necessary (can do a map lookup with only one value on LHS if we're not using the second bool)
</blockquote></details>

Done.

nmiyake (approved):
> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/105)* status: all files reviewed, 1 unresolved discussion (waiting on @k-simons)

___
*[conjure/serverwriter.go, line 701 at r2](https://reviewable.io/reviews/palantir/conjure-go/105#-Lzh7zR18qTDGOccRRl8:-LzhJHa1Ej-BBs0q86AM:b-nf5o6r) ([raw file](https://github.com/palantir/conjure-go/blob/14cd38971769cb4e2d83bb1819f2b9f402d8d953/conjure/serverwriter.go#L701)):*
<details><summary><i>Previously, k-simons wrote…</i></summary><blockquote>

So at a high level, I think that you would want to refactor `ParseStringParam` into just being `ParseParam` and then correctly use the internal visitor that it has. 

I poked at it and it was more effort than I was willing to give
</blockquote></details>

Chatted offline and OK with going through with this as-is -- I'll likely add support for the rest of the spec here shortly, but won't block this PR on it.

PR #106, opened 2020-01-29
Improvement: update Conjure API definition to 4.7.0

nmiyake:
## Before this PR
Repository was not using latest version of Conjure API definitions, and it was hard to tell the version of the API definitions being used.

## After this PR
==COMMIT_MSG==
Uses the latest Conjure API definitions and updates generation file code and IR filename to include the version.
==COMMIT_MSG==

## Possible downsides?
Possible that there are some features/definitions in newest API definitions that are not fully supported by `conjure-go`, but at a minimum none of the testing broke, so I don't think this is a major risk.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/106)


conjure-api/generate.go
```diff
@@ -16,4 +16,4 @@
 
 package conjure
 
-//go:generate go run ../main.go conjure-api.json
+//go:generate go run ../main.go conjure-api-4.7.0.conjure.json
```
tabboud, line 19: 👍

tabboud (approved):

PR #114, opened 2020-05-04
Improvement: Generated errors implement ParamStorer interface

asanderson15:
## Before this PR
There was no way to know which param fields in the generated error types were safe vs unsafe.

## After this PR
==COMMIT_MSG==
Generated error types now implement the [`wparams.ParamStorer`](https://github.com/palantir/witchcraft-go-params/blob/develop/paramstorer.go) interface. This ensures any downstream consumers of the error (e.g., witchcraft-go-logging) will automatically do the right thing with the params.
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/114)

### DIFF
--- a/conjure/errorwriter.go
+++ b/conjure/errorwriter.go
@@ -33,6 +33,7 @@ import (
 const (
 	errorReceiverName    = "e"
 	errorInstanceIDField = "errorInstanceID"
+	errorInstanceIDParam = "errorInstanceId"
 )
 
 const (
@@ -359,7 +360,7 @@ func astErrorSafeParamsMethod(errorDefinition spec.ErrorDefinition, info types.P
 		))
 	}
 	keyValues = append(keyValues, expression.NewKeyValue(
-		fmt.Sprintf("%q", errorInstanceIDField),
+		fmt.Sprintf("%q", errorInstanceIDParam),
 		expression.NewSelector(
 			expression.VariableVal(errorReceiverName),
 			errorInstanceIDField,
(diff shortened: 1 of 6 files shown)


conjure/errorwriter.go
```diff
@@ -340,6 +342,84 @@ func astErrorParametersMethod(errorDefinition spec.ErrorDefinition, info types.P
...
 }
 
+// astErrorSafeParamsMethod generates SafeParams function for an error, for example:
+//
+//  func (e *MyNotFound) SafeParams() map[string]interface{} {
+//  	return map[string]interface{}{"safeArgA": e.safeArgA, "safeArgB": e.safeArgB}
+//  }
+func astErrorSafeParamsMethod(errorDefinition spec.ErrorDefinition, info types.PkgInfo) astgen.ASTDecl {
+	var keyValues []astgen.ASTExpr
+	for _, fieldDefinition := range errorDefinition.SafeArgs {
```
bmoylan, line 353: should we add errorInstanceId as an extra one here?

integration_test/testgenerated/errors/errors_test.go
```diff
@@ -83,3 +83,17 @@ func TestError_UnmarshalJSON(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, testError, &myNotFound)
 }
+
```
bmoylan, line 88: can we add a quick type check test to cast to ParamStorer? Something like
```go
func TestEror_isParamStorer(t *testing.T) {
	_ = testError.(wparams.ParamStorer)
}
```

conjure/errorwriter.go
asanderson15: Yes, we definitely should. Good catch.

bmoylan (approved):

PR #115, opened 2020-05-05
Improvement: Register error types with CGR via init fn

asanderson15:
## Before this PR
CGR has no way to match a conjure defined server error to its generated type.

## After this PR
==COMMIT_MSG==
Each generated conjure error type is registered with CGR's error registry via an `init()` function, allowing unmarshal errors into the generated error struct types. This unblocks returning real conjure errors. One init function will be added to register all errors in a given package. 
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/115)


bmoylan (commented):
> Reviewed 19 of 19 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/115)* status: all files reviewed, 2 unresolved discussions (waiting on @asanderson15)

___
*[conjure/conjure_test.go, line 2217 at r1](https://reviewable.io/reviews/palantir/conjure-go/115#-M6aS7Qe6ho_zHbAUXFN:-M6aS7Qe6ho_zHbAUXFO:b-m9p4mn) ([raw file](https://github.com/palantir/conjure-go/blob/38b54353c908acdf66cc7740568ef9cc9585e6cc/conjure/conjure_test.go#L2217)):*
> ```Go
> 
> func init() {
> 	errors.RegisterErrorType("MyNamespace:MyNotFound", reflect.TypeOf(MyNotFound{}))
> ```

can we add a second error definition to this test (or a new one) so we can see them aggregated here?
___
*[integration_test/testgenerated/errors/errors_test.go, line 109 at r1](https://reviewable.io/reviews/palantir/conjure-go/115#-M6aSKRA6O0NvYpbnGqe:-M6aSKRB3AFekWKcmcAm:bplwwu8) ([raw file](https://github.com/palantir/conjure-go/blob/38b54353c908acdf66cc7740568ef9cc9585e6cc/integration_test/testgenerated/errors/errors_test.go#L109)):*
> ```Go
> 	myNotFoundErr, ok := genericErr.(*api.MyNotFound)
> 	require.True(t, ok)
> 	assert.Equal(t, myNotFoundErr.UnsafeArgA, testError.UnsafeArgA)
> ```

Can the full structs be compared? like `assert.Equal(t, myNotFoundErr, testError)`

bmoylan (commented):
LGTM but will hold off on merging till we can deal with the dependency dance

> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/115#-:-M6aq_FR80XvJZeLXgOw:bnqt1ik)* status: all files reviewed, 1 unresolved discussion (waiting on @asanderson15)

bmoylan (approved):

PR #116, opened 2020-05-07
Break: Bump conjure-go-runtime to v2

asanderson15:
## After this PR
==COMMIT_MSG==
Bump conjure-go-runtime to v2 to pick up improved conjure typed error handling.
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/116)


integration_test/testgenerated/auth/auth_test.go
```diff
@@ -62,7 +62,7 @@ func TestBothAuthClient(t *testing.T) {
 
 	// test invalid auth
 	_, err = client.Default(ctx, "invalid token")
-	assert.EqualError(t, err, "httpclient request failed: server returned a status >= 400")
+	assert.EqualError(t, err, "httpclient request failed: failed to unmarshal body as conjure error: json: cannot unmarshal string into Go value of type struct { Name string \"json:\\\"errorName\\\"\" }")
```
bmoylan, line 65: not going to block on it, but something's fishy about this -- this either means we're returning JSON which doesn't match conjure or we're claiming it's application/json when it isn't

bmoylan (approved):

go.sum
```diff
@@ -98,6 +98,8 @@ github.com/openzipkin/zipkin-go v0.2.2 h1:nY8Hti+WKaP0cRsSeQ026wU03QsM762XBeCXBb
 github.com/openzipkin/zipkin-go v0.2.2/go.mod h1:NaW6tEwdmWMaCDZzg8sh+IBNOxHMPnhQw8ySjnjRyN4=
 github.com/palantir/conjure-go-runtime v0.4.0 h1:xmSzYVv27wnKNCcmtamcO9XpI6+Yig/cNElomWU72f4=
```
asanderson15, line 99: Is it expected that this wouldn't get cleaned up?

integration_test/testgenerated/auth/auth_test.go
asanderson15: Yeah, looks like https://github.com/palantir/witchcraft-go-server/blob/352f44636377416a16468535e8d8847e2137294d/rest/rest.go#L28 always writes that header.

PR #124, opened 2020-05-19
Fix error constructors with fields that are type names

### LINKED ISSUE #96: Error parameter with the same name as the error causes compile breaks
## What happened?
Defining an error with a parameter that has the same name as the error cause a break in generated code, example below:

```
InvalidRids:
  namespace: Example
  code: INVALID_ARGUMENT
  docs: |
    Thrown when an operation is attempted on invalid RIDs
  safe-args:
    invalidRids: set<rid>
```

This generates code that looks like this:
```
// NewInvalidRids returns new instance of InvalidRids error.
func NewInvalidRids(invalidRids []rid.ResourceIdentifier) *InvalidRids {
	return &InvalidRids{errorInstanceID: uuid.NewUUID(), invalidRids: invalidRids{InvalidRids: foo}}
}

// InvalidRids is an error type.
//
// Thrown when an operation is attempted on invalid RIDs
type InvalidRids struct {
	errorInstanceID uuid.UUID
	invalidRids
}
```

Would prefer if the generated code renamed the parameter to avoid the compile break.

bmoylan:
## Before this PR
If an error used a parameter name that was also a type name, the constructor's function argument shadowed the type and the code does not compile.
## After this PR
==COMMIT_MSG==
Fix error constructors with fields that are type names
==COMMIT_MSG==

Fixes #96 

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/124)

### DIFF
--- a/changelog/@unreleased/pr-124.v2.yml
+++ b/changelog/@unreleased/pr-124.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Fix error constructors with fields that are type names
+  links:
+  - https://github.com/palantir/conjure-go/pull/124
--- a/conjure/conjure_test.go
+++ b/conjure/conjure_test.go
@@ -2145,8 +2145,8 @@ func (o *myNotFound) UnmarshalYAML(unmarshal func(interface{}) error) error {
 }
 
 // NewMyNotFound returns new instance of MyNotFound error.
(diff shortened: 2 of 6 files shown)


nmiyake (approved):
> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/124)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #130, opened 2020-06-25
Fix client sending list/set query params

### LINKED ISSUE #128: Bad request response for lists and sets as query parameters
## What happened?

A list or set `somefield` as URL parameter is translated into a URL string  `somefield=%5Bvalue1,value2%5D`. This format is not supported by all servers, and in our specific case we get an HTTP 400 response.

Example for such a field:
```
{
    "fieldName": "somefield",
    "type": {
        "type": "optional",
        "optional": {
            "itemType": {
                "type": "set",
                "set": {
                    "itemType": {
                        "type": "external",
                        "external": {
                            "externalReference": {
                                "name": "Somefield",
                                "package": "com.somepackage"
                            },
                            "fallback": {
                                "type": "primitive",
                                "primitive": "STRING"
                            }
                        }
                    }
                }
            }
        }
    },
    "docs": null
}
```

It seems that a translation like `somefield=value1&somefield=value2` would be supported more widely (in our case this is the only format that our server accepts). This seems not to be supported by URL translation via Encode in the net/url package (as used in github.com/palantir/conjure-go-runtime/v2@v2.2.0/conjure-go-client/httpclient/client.go) though.

bmoylan:
## Before this PR
Fixes #128: Bad request response for lists and sets as query parameters

## After this PR
==COMMIT_MSG==
Generated clients correctly set list/set query parameters as multiple values
==COMMIT_MSG==

## Possible downsides?
Technically a break to how clients send data over the wire, though I'm marking it as a bugfix since we were not in compliance with the spec.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/130)

### DIFF
--- a/changelog/@unreleased/pr-130.v2.yml
+++ b/changelog/@unreleased/pr-130.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Generated clients correctly set list/set query parameters as multiple
+    values
+  links:
+  - https://github.com/palantir/conjure-go/pull/130
--- a/conjure/servicewriter.go
+++ b/conjure/servicewriter.go
@@ -726,23 +726,43 @@ func serviceStructMethodBodyAST(endpointDefinition spec.EndpointDefinition, retu
 		for _, queryParam := range queryParams {
 			currQueryParamVarName := argNameTransform(string(queryParam.ArgumentDefinition.ArgName))
 			currQueryParamKeyName := visitors.GetParamID(queryParam.ArgumentDefinition)
-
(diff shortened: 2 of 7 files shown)


nmiyake (approved):
Thanks!

> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/130#-:-MAhEH-KCpoFjYVvABON:b-ucfknh)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #136, opened 2020-08-25
use boolean.Boolean for bool map keys

tabboud:
## Before this PR
bool map keys were not wire compatible, since the conjure spec uses the string representation for all map keys, which resulted in conjure-go clients being unable to deserialize requests with bool map keys.
For example:
```yml
types:
  definitions:
    default-package: com.palantir.test.api
    objects:
      Result:
        fields:
          counts: map<boolean, integer>
```
The above definition would produce the following struct type:
```go
type Result struct {
	Counts map[bool]int `json:"counts"`
}
```
The expected serialized form for the above should be the following:
```json
{
    "counts": {
        "false": 200,
        "true": 100
    }
}
```

However go will fail to unmarshal the string form of a bool into the bool type, which is why we need to use the `boolean.Boolean` type instead.

## After this PR

==COMMIT_MSG==
use the new boolean.Boolean type for bool map keys
==COMMIT_MSG==

## Possible downsides?
This breaks the generated API since we no longer use the primitive bool type and instead use the new `boolean.Boolean` type. The previous behavior was technically incorrect anyways since clients were unable to serialize/deserialize requests correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/136)


bmoylan (approved):
Looks good! Needs changelog

> Reviewed 16 of 16 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/136#-:-MFbAJTe2BcWyVdEpUCq:b-91u0ik)* status: :shipit: complete! all files reviewed, all discussions resolved

bmoylan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/136)* status: all files reviewed, 1 unresolved discussion (waiting on @tabboud)

*[a discussion](https://reviewable.io/reviews/palantir/conjure-go/136#-MFbAVcT6ew6F_2fmB3s:-MFbAVcT6ew6F_2fmB3t:b7yw5za) (no related file):*
Actually, can you add to one of the integration tests (testgenerated/*) to make sure that the ser/deser works?

---

bmoylan:
*[a discussion](https://reviewable.io/reviews/palantir/conjure-go/136#-MFbAVcT6ew6F_2fmB3s:-MFbAftk70kcqp20FFak:b-8zo1qj) (no related file):*

Actually, can you add to one of the integration tests (testgenerated/*) to make sure that the ser/deser works?
</blockquote></details>

Well, I guess the verifier kind of does that already

---

tabboud:
The verifier doesn't test the full serde, so I went ahead and added a few more cases.
Also for the changelog, should we treat this as a break since the generated code will be different, or is this purely a fix since the behavior before was incorrect?

bmoylan (approved):
I'm ok with calling it a `fix` since the previous implementation would not have worked on the wire and in most cases the compiler will catch any issues arising from the generated code change.

> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/136#-:-MFbWUQc3RrO5NGnJMQu:b-iu884r)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #138, opened 2020-09-13
Server route registration recognizes com.palantir.logsafe.Safe argument marker

### LINKED ISSUE #121: Allow for marking endpoint arguments Safe
Java has a `@Safe` annotation for marking params as safe which is attached via `markers`, we should figure out a way to do the same. This requirement will make it tough since this seems to be overfit to java annotations:
> List of types that serve as additional metadata for the argument. If the value of the field is a string it MUST be a type name that exists within the Conjure definition.

bmoylan:
## Before this PR
There was no way to enable safe logging on server parameters. See some discussion on https://github.com/palantir/conjure/issues/612.

## After this PR

==COMMIT_MSG==
Server route registration recognizes `com.palantir.logsafe.Safe` argument marker
==COMMIT_MSG==

Fixes #121 

## Possible downsides?
Hardcoding this is obviously brittle, though it matches what conjure-rust does and is probably worth it given this API is so widely used in the rest of the conjure ecosystem.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/138)

### DIFF
--- a/integration_test/testgenerated/server/server-service.yml
+++ b/integration_test/testgenerated/server/server-service.yml
@@ -1,4 +1,8 @@
 types:
+  imports:
+    Safe:
+      external:
+        java: com.palantir.logsafe.Safe
   definitions:
     default-package: api
     objects:
@@ -122,6 +126,50 @@ services:
             param-type: header
             param-id: X-My-Header2
         returns: CustomObject
+        #markers:
+        #              - Safe
+      postSafeParams:
+        http: POST /safe/{myPathParam1}/{myPathParam2}
+        auth: header
+        args:
+          myPathParam1:
+            type: string
+            markers:
+              - Safe
+          myPathParam2: boolean
+          myBodyParam:
+            type: CustomObject
+            param-type: body
+          myQueryParam1:
+            type: string
+            param-type: query
+            param-id: query1
+            markers:
+              - Safe
+          myQueryParam2:
+            type: string
+            param-type: query
+            markers:
+              - Safe
+          myQueryParam3:
+            type: double
+            param-type: query
+          myQueryParam4:
+            type: optional<safelong>
+            param-type: query
+          myQueryParam5:
+            type: optional<string>
+            param-type: query
+          myHeaderParam1:
+            type: safelong
+            param-type: header
+            param-id: X-My-Header1-Abc
+            markers:
+              - Safe
+          myHeaderParam2:
+            type: optional<uuid>
+            param-type: header
+            param-id: X-My-Header2
       bytes:
         http: GET /bytes
         returns: CustomObject
--- a/integration_test/testgenerated/server/server-service.yml
+++ b/integration_test/testgenerated/server/server-service.yml
@@ -3,6 +3,9 @@ types:
     Safe:
       external:
         java: com.palantir.logsafe.Safe
+    OtherMarker:
+      external:
+        java: com.palantir.logsafe.OtherMarker
   definitions:
     default-package: api
     objects:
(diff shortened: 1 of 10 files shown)


bmoylan:
friendly ping @tabboud

integration_test/testgenerated/server/server_test.go
```diff
@@ -0,0 +1,174 @@
...
+	myQueryParam3Arg float64, myQueryParam4Arg *safelong.SafeLong, myQueryParam5Arg *string, myHeaderParam1Arg safelong.SafeLong, myHeaderParam2Arg *uuid.UUID) error {
+	switch {
+	case authHeader == "":
+		return errors.New("empty authHeader")
+	case myPathParam1Arg == "":
+		return errors.New("empty myPathParam1Arg")
+	case myPathParam2Arg == false:
+		return errors.New("empty myPathParam2Arg")
+	case reflect.ValueOf(myBodyParamArg).IsZero():
+		return errors.New("empty myPathParam1Arg")
```
tabboud: `s/myPathParam1Arg/myBodyParamArg`

integration_test/testgenerated/server/server-service.yml
```diff
@@ -122,6 +126,50 @@ services:
             param-type: header
             param-id: X-My-Header2
         returns: CustomObject
+        #markers:
```
tabboud: remove

integration_test/testgenerated/server/server-service.yml
```diff
@@ -1,4 +1,8 @@
 types:
+  imports:
+    Safe:
+      external:
+        java: com.palantir.logsafe.Safe
```
tabboud, line 5: There is already a unit test for the invalid import, but it might be nice to add another import with a different path and use that as a marker below to ensure it doesn't get processed as safe.

integration_test/testgenerated/server/server-service.yml
```diff
@@ -15,16 +19,16 @@ services:
...
-        auth: header
-        args:
-          myPathParam: string
       getPathParamAlias:
         http: GET /path/alias/{myPathParam}
         auth: header
         args:
           myPathParam: StringAlias
+      getPathParam:
+        http: GET /path/string/{myPathParam}
```
tabboud, line 26: Whats the reason for adding in the `/string/` to the path?

integration_test/testgenerated/server/server_test.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/palantir/conjure-go/v5/integration_test/testgenerated/server/api"
+)
+
+func TestSafeMarker(t *testing.T) {
+	router := wrouter.New(whttprouter.New())
+	err := api.RegisterRoutesTestService(router, testServerImpl{})
+	require.NoError(t, err)
+	called := false
+	router.AddRouteHandlerMiddleware(func(rw http.ResponseWriter, r *http.Request, reqVals wrouter.RequestVals, next wrouter.RouteRequestHandler) {
+		if reqVals.Spec.PathTemplate == "/safe/{myPathParam1}/{myPathParam2}" {
```
tabboud, line 31: note: It would be nice if we generated these as const values in the server file so we wouldn't have to copy+paste. This comes up anytime we need to implement route middleware.

tabboud (changes requested):

integration_test/testgenerated/server/server-service.yml
bmoylan: Yeah this is kind of lame. It turns out this was not working before, but we never actually registered the routes with httprouter and hit this panic when we do: https://github.com/julienschmidt/httprouter/blob/v1.3.0/tree.go#L163-L167

```
panic: 'alias' in new path '/path/alias/:myPathParam' conflicts with existing wildcard ':myPathParam' in existing prefix '/path/:myPathParam'
```

I'm not sure if this is a w-g-s bug and we should be handling these...

tabboud: This appears to be an issue specific to `julienschmidt/httprouter` since it uses prefix matching on the routes. Since there can only be 1 path to 1 route, it doesn't know which route to choose here since `alias/:name` could be matched for the same `/path/:name` portion. It seems like adding in the static path segment `string` resolves the common prefixes.

tabboud (approved):

Issue #24, opened 2019-01-14
Generated code does not support round-tripping of unknown enum/union types


nmiyake:
The Conjure specification stipulates that clients should be able to round-trip unknown variants of enums and unions: https://github.com/palantir/conjure/blob/master/docs/spec/wire.md#44-round-trip-of-unknown-variants

This means that, even if a union unmarshals into an "unknown" type, it should be able to re-marshal the unknown union into its original form.

Currently, the generated code does not do this -- although it will unmarshal unknown variants, the serializer code returns an error when an unmarshaled unknown is re-marshaled.

In order to fix this, the generated unions will likely need to store the raw JSON and re-write that when it is marshaled. Not sure if there's a straightforward way to properly support this for YAML, but at the very least we should match the required JSON behavior.

PR #139, opened 2020-09-15
Enums roundtrip unknown variants

bmoylan:
## Before this PR
Unknown enum variants were all assigned the `UNKNOWN` value, which is not in line with the conjure spec that specifies they should be stored and round-trippable. Partially addresses #24 (unions require more thought).

## After this PR

==COMMIT_MSG==
Enums now preserve unknown variant values. The `UNKNOWN` value has been removed.
==COMMIT_MSG==

## Possible downsides?

This might be a breaking change, especially if people are assuming the generated values are exhaustive (e.g. using switch statements without default cases)

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/139)


conjure/enum.go
```diff
@@ -131,3 +135,27 @@ func enumUnmarshalTextAST(e spec.EnumDefinition, info types.PkgInfo) astgen.ASTD
 	}
 	return newUnmarshalTextMethod(enumReceiverName, transforms.Export(e.TypeName.Name), switchStmt, statement.NewReturn(expression.Nil))
 }
+
+// Goastwriter does not allow values in var declarations, so implement it here.
+// This should be contributed back to goastwriter.
```
bmoylan: https://github.com/palantir/goastwriter/pull/58

tabboud: I went ahead and approved the goastwriter change above, so feel free to rebase and I can review this one.

bmoylan: Done!

asanderson15 (approved):

asanderson15 (changes requested):
Test is broken. Options are to change the test code not to throw the warning or just bump verifier to 0.16.6, which has the new correct enum test cases.

conjure/enum.go
```diff
@@ -54,32 +56,68 @@ func astForEnum(enumDefinition spec.EnumDefinition, info types.PkgInfo) []astgen
...
+					&statement.If{
+						Cond: expression.NewUnary(token.NOT,
+							expression.NewCallExpression(
+								expression.NewSelector(expression.VariableVal(enumPatternVarName), "MatchString"),
+								expression.VariableVal(enumUpperVarName),
+							),
+						),
+						Body: []astgen.ASTStmt{
+							statement.NewReturn(
+								expression.NewCallFunction("werror", "Convert",
```
tabboud, line 100: What is the value in adding in the additional `werror.Convert` given the `NewInvalidArgument` will return a proper error?

bmoylan: Adam asked the same question on Slack. This might be exposing a deficiency with how we implemented the conjure error, but because it's not a werror it doesn't have a stacktrace. I added the werror.Convert as low as possible to preserve the stack frame(s) between here and the next werror.Wrap.

asanderson15 (approved):

tabboud (approved):

PR #151, opened 2020-10-06
Improvement: Generate IsErrorType helpers for conjure errors

### LINKED ISSUE #141: Conjure errors should have comparison helper function
## What happened?

A common use case is to look at a returned error and determine whether it's a known type and handle accordingly. E.g.:

```
err := doSomething()
if err != nil {
    if e, ok := werror.RootCause(err).(errors.Error); ok {
        if e.Name() == "Namespace:FooBarConjureError" {
            // do something
        }
    }
}
```

That's not great because (a) it's a lot of boilerplate and (b) it requires copy-pasting the error name from the generated conjure since it's not exported as a const.

Instead, we should generate a small helper function:

```
func IsFooBarConjureError(err error) bool {
    if err == nil {
        return false
    }
    if e, ok := werror.RootCause(err).(errors.Error); ok {
        if e.Name() == "Namespace:FooBarConjureError" {
           return true
        }
    }
    return false
}
```

This would turn the above code into:

```
err := doSomething()
if err != nil && api.IsFooBarConjureError(err) {
    // do something
}
```

asanderson15:
==COMMIT_MSG==
Generate IsErrorType helpers for conjure errors
==COMMIT_MSG==

Fixes #141

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/151)

### DIFF
--- a/changelog/@unreleased/pr-151.v2.yml
+++ b/changelog/@unreleased/pr-151.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Generate IsErrorType helpers for conjure errors
+  links:
+  - https://github.com/palantir/conjure-go/pull/151
--- a/conjure/conjure_test.go
+++ b/conjure/conjure_test.go
@@ -2163,6 +2163,7 @@ import (
 	"github.com/palantir/pkg/safejson"
 	"github.com/palantir/pkg/safeyaml"
 	"github.com/palantir/pkg/uuid"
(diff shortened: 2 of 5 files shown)


bmoylan (approved):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/151)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tabboud)

PR #167, opened 2020-11-11
Render servers with witchcraft-go-server v2

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/167)


asanderson15 (approved):

nmiyake (changes requested):
> Reviewed 2 of 86 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/167)* status: 2 of 86 files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[changelog/@unreleased/pr-167.v2.yml, line 1 at r1](https://reviewable.io/reviews/palantir/conjure-go/167#-MLomHmy3MGVlOuN1n9O:-MLomHmy3MGVlOuN1n9P:bdtlawt) ([raw file](https://github.com/palantir/conjure-go/blob/463dcb5b16086137379bfb9358748e411b37b5db/changelog/@unreleased/pr-167.v2.yml#L1)):*
> ```YAML
> type: fix
> ```

Is the server generation still considered experimental (although not sure that really has a practical meaning since I think we have a lot of projects using it now)? Will this change cause new code to be written that will break previously compiling things? If so, should this be considered a break?

Even if we don't consider it a break, "feature" or "improvement" seems like a better fit than "fix" for the scope of this change.

bmoylan:
___
*[changelog/@unreleased/pr-167.v2.yml, line 1 at r1](https://reviewable.io/reviews/palantir/conjure-go/167#-MLomHmy3MGVlOuN1n9O:-MLooTvN3M31X1GdJZup:b-d1eolf) ([raw file](https://github.com/palantir/conjure-go/blob/463dcb5b16086137379bfb9358748e411b37b5db/changelog/@unreleased/pr-167.v2.yml#L1)):*

Is the server generation still considered experimental (although not sure that really has a practical meaning since I think we have a lot of projects using it now)? Will this change cause new code to be written that will break previously compiling things? If so, should this be considered a break?

Even if we don't consider it a break, "feature" or "improvement" seems like a better fit than "fix" for the scope of this change.
</blockquote></details>

So, you caught us. The intention of the `fix` was to sneak this into the 6.x breaking changes after discovering that we can't excavate out the WGS V2 change until conjure-generated servers accept the new wrouter import. Alternatively we could do another break and go to 7.x, or call this a `break` in the changelog but hold the version back to 6.0.1.

nmiyake (approved):
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/167)* status: 2 of 86 files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[changelog/@unreleased/pr-167.v2.yml, line 1 at r1](https://reviewable.io/reviews/palantir/conjure-go/167#-MLomHmy3MGVlOuN1n9O:-MLorKYa52i1zJiHYHQZ:b-jti49) ([raw file](https://github.com/palantir/conjure-go/blob/463dcb5b16086137379bfb9358748e411b37b5db/changelog/@unreleased/pr-167.v2.yml#L1)):*

So, you caught us. The intention of the `fix` was to sneak this into the 6.x breaking changes after discovering that we can't excavate out the WGS V2 change until conjure-generated servers accept the new wrouter import. Alternatively we could do another break and go to 7.x, or call this a `break` in the changelog but hold the version back to 6.0.1.
</blockquote></details>

Ah OK -- I don't mind rolling in as part of the v6 release (especially if doing another bump would require another round of bumping across repos, since then the cost likely isn't worth it), but in that case would just like this context included in the changelog -- basically, note that this includes a breaking change that was meant to be part of the initial v6 release and update the v6 release notes to note that that version shouldn't be used.

Will change to approve to unblock the merge if you want to proceed, but just want to make sure that we can reason about what happened later.

bmoylan:
___
*[changelog/@unreleased/pr-167.v2.yml, line 1 at r1](https://reviewable.io/reviews/palantir/conjure-go/167#-MLomHmy3MGVlOuN1n9O:-MLoyc841O69iLXdRWIh:blvv0qd) ([raw file](https://github.com/palantir/conjure-go/blob/463dcb5b16086137379bfb9358748e411b37b5db/changelog/@unreleased/pr-167.v2.yml#L1)):*

Ah OK -- I don't mind rolling in as part of the v6 release (especially if doing another bump would require another round of bumping across repos, since then the cost likely isn't worth it), but in that case would just like this context included in the changelog -- basically, note that this includes a breaking change that was meant to be part of the initial v6 release and update the v6 release notes to note that that version shouldn't be used.

Will change to approve to unblock the merge if you want to proceed, but just want to make sure that we can reason about what happened later.
</blockquote></details>

Thanks! I changed the type to a break and added some additional language about the change. Let me know if you think this is clear enough. After this merges I'll make sure to force the release to 6.0.1

PR #169, opened 2020-11-12
Fix: Don't fail deserialization on invalid enum values

asanderson15:
## Before this PR
When enum roundtripping was added, a constraint was also added to deserialization that would throw an error if a value failed to meet the constraints applied when generating conjure IR. However, this was inconsistent with [conjure-java](https://github.com/palantir/conjure-java/blob/6816a503a153840a3a9949db4087bcf2366c2045/conjure-java-core/src/integrationInput/java/test/prefix/com/palantir/product/SimpleEnum.java#L60-L70), and was stricter than the conjure spec requires.

## After this PR

==COMMIT_MSG==
Don't fail deserialization on invalid enum values, making behavior consistent with conjure-java.
==COMMIT_MSG==

By removing this constraint, conjure-go's behavior matches conjure-java.

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/169)

### DIFF
--- a/conjure-go-verifier/ignored-test-cases.yml
+++ b/conjure-go-verifier/ignored-test-cases.yml
@@ -111,6 +111,9 @@ client:
     - "null"
     receiveSetDoubleAliasExample:
     - "[100, 10.0, \"NaN\", \"Infinity\", \"-Infinity\"]"
+    receiveEnumExample:
+    - "\"!!!\""
+    - "\"one-hundred\""
   singlePathParamService:
     pathParamAliasString:
     - '""'
--- a/changelog/@unreleased/pr-169.v2.yml
+++ b/changelog/@unreleased/pr-169.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Don't fail deserialization on invalid enum values, making behavior
+    consistent with conjure-java.
+  links:
+  - https://github.com/palantir/conjure-go/pull/169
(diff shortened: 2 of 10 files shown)


conjure-go-verifier/ignored-test-cases.yml
```diff
@@ -111,6 +111,9 @@ client:
     - "null"
     receiveSetDoubleAliasExample:
     - "[100, 10.0, \"NaN\", \"Infinity\", \"-Infinity\"]"
+    receiveEnumExample:
```
asanderson15, line 114: Note that these two cases are similarly ignored by conjure-java as well: https://github.com/palantir/conjure-java/blob/c9de473727fe1e685c3ad7e24f3c2023c3e7f009/conjure-java-client-verifier/src/test/resources/ignored-test-cases.yml#L107-L109.

nmiyake (approved):
OK -- basically idea is that invalid generation should be prevented by IR checker so by the time we unmarshal, if it's invalid we should just treat it as a "normal" invalid value, correct?

Assuming that the above is correct, this :lgtm: 

> Reviewed 10 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/169#-:-MLyyLf_Brhh_xeKW8kE:b8wgtui)* status: all files reviewed, 1 unresolved discussion (waiting on @bmoylan)

asanderson15:
That's correct.

PR #185, opened 2021-02-04
Generate product dependencies information from IR for services

nmiyake:
## Before this PR
If input Conjure IR contained product dependency information in its "Extensions" block, nothing was done with it.

## After this PR

==COMMIT_MSG==
If the Conjure IR used for code generation contains product dependency
information in its extensions block and defines any services, then a
"zz_generated_product_dependencies.go" file that contains the product
dependency information is written in the same package as the generated
service Go files. This makes it so that product dependency information
for the generated Conjure can be read by other products.
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/185)


nmiyake:
@kazimierzbudzyk @smoorpal for SA

bmoylan (approved):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/185)* status: :shipit: complete! all files reviewed, all discussions resolved

bmoylan:
@nmiyake is this still something we want to get in?

nmiyake:
Thanks for the bump -- there's enough distance from this now that I don't think we should proceed as is unless/until we revive the broader project. Will close for now.

PR #190, opened 2021-03-02
Improvement: add AcceptFuncs lambda visitor

jhowarth-pal:
Implementation of Proposal 1 from https://palantir.quip.com/XPyEApEdaBqb

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/190)


changelog/@unreleased/pr-190.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: add AcceptFuncs lambda visitor
```
nmiyake: Changelogs are meant to be consumed by end users as part of the release notes, so this description should be more extensive -- as it stands, I wouldn't be able to tell what this means if I were to just read the release notes without further context. The heading/first sentence can be a summary, but that should be followed with more details that explains what was added here in a manner that is understandable to someone reading the release notes.

As an example:

```
Generate an AcceptFuncs lambda visitor for union types.

Adds an option to generate an "AcceptFuncs" function for union types which accepts lambda functions that can be used as visitors. This allows visitors to be called on union types without having to implement a visitor interface.
```

Doesn't have to be exactly the above, but hopefully that provides an idea.

conjure-api/conjure/spec/unions.conjure.go
```diff
@@ -76,6 +76,20 @@ func (u *AuthType) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return safejson.Unmarshal(jsonBytes, *&u)
 }
 
+func (u *AuthType) AcceptFuncs(headerFunc func(HeaderAuthType) error, cookieFunc func(CookieAuthType) error, unknownFunc func(string) error) error {
+	switch u.typ {
+	default:
+		if u.typ == "" {
+			return fmt.Errorf("invalid value in union type")
```
nmiyake: Is there a reason that that `""` is special-cased here rather than passing it to `unknownFunc`?

nmiyake (changes requested):
Minor feedback on changelog content and handling of `""` for unknown types, but otherwise general implementation looks good!

conjure-api/conjure/spec/unions.conjure.go
jhowarth-pal: I was following the precedent from `Accept` [below](https://github.com/palantir/conjure-go/pull/190/files#diff-6901295dd2f0f080dfaeb8e7d1600d638438e7db8a8e5adebd9e8f5e5a5e9fd3R96)

I defer to your opinion here, but I think we should be consistent between `Accept` and `AcceptFuncs` either way

changelog/@unreleased/pr-190.v2.yml
jhowarth-pal: Thanks, good catch

conjure-api/conjure/spec/unions.conjure.go
nmiyake: Ah thanks for the reference -- yes agree that should be consistent and this is fine.

nmiyake (commented):
Thanks, core implementation here looks good! Next steps would be to add a flag to control this generation (much like server generation) and to add integration tests (including testing the generated code by adding cases in `integration_test/testgenerated/objects` or making a similar new package)

jhowarth-pal:
@nmiyake added the feature flag and exercised the code paths within existing integration tests.

I just used the previously defined visitor methods as the functions themselves since that tests the code paths. This seemed sufficient to me, but I'm happy to add more if you'd like

nmiyake (approved):

PR #191, opened 2021-03-15
Improvement: support optional binary

### LINKED ISSUE #192: empty optional<binary> should return as a 204 status code
Per the [conjure spec](https://github.com/palantir/conjure/blob/master/docs/spec/wire.md#31-status-codes), an endpoint that returns an `optional<binary>` type should return an empty stream with a status code of `204`. The distinguishes an empty binary from a present binary of zero length. 

I attempted to take advantage of this by specifying an endpoint with an `optional<binary>` type and returning nil in the resource layer if the binary was not found. Rather than return a `204` status code I got a `500` status code. One key issue is that [this special casing](https://github.com/palantir/conjure-go/blob/develop/conjure/servicewriter.go#L1026) treats `optional<binary>` and `binary` identically so the conjure go infra fails on a nil return type.

The solution is to improve the special casing to convert `optional<binary>` -> `*ReadCloser` rather than the standard `ReadCloser` object and providing handling for the `nil` -> `204` case.

AdamDeHovitz:
## Before this PR
`optional<binary>` equivalent was getting mapped in go to `ReadCloser` type rather than `*ReadCloser` due to flawed special casing.

Fixes: https://github.com/palantir/conjure-go/issues/192

## After this PR

==COMMIT_MSG==
`optional<binary>` now represented in go as `*ReadCloser` and a 204 status code is treated as nil
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/191)


AdamDeHovitz:
~On second thought this would also convert `set<binary>` to `set<ReadCloser>` which may not be valid.~

conjure/conjure_test.go
```diff
@@ -1103,6 +1123,20 @@ func (c *testServiceClient) StreamResponse(ctx context.Context, authHeader beare
...
+	requestParams = append(requestParams, httpclient.WithRPCMethodName("MaybeStreamResponse"))
+	requestParams = append(requestParams, httpclient.WithRequestMethod("GET"))
+	requestParams = append(requestParams, httpclient.WithHeader("Authorization", fmt.Sprint("Bearer ", authHeader)))
+	requestParams = append(requestParams, httpclient.WithPathf("/catalog/maybe/streamResponse"))
+	requestParams = append(requestParams, httpclient.WithRawResponseBody())
+	resp, err := c.client.Do(ctx, requestParams...)
+	if err != nil {
+		return nil, err
+	}
+	return resp.Body, nil
```
nmiyake: Since this is the client code, I think the logic we actually want is:

```
if resp.StatusCode == http.StatusNoContent {
        return nil, nil
}
```

Then the contract for these functions is that a return value of `nil` for the `ReadCloser` indicates that there was no content (pretty sure that if the response is empty, then `resp.Body` will be non-nil but will return 0 bytes when read, which is the correct behavior)

conjure/servicewriter.go
```diff
@@ -1018,25 +1018,13 @@ func returnTypesForEndpoint(endpointDefinition spec.EndpointDefinition, info typ
...
-		returnBinary, err := isReturnTypeSpecificType(endpointDefinition.Returns, visitors.IsBinary)
+		typer, err := visitors.NewConjureTypeProviderTyper(*endpointDefinition.Returns, info)
 		if err != nil {
 			return nil, nil, err
 		}
-		if returnBinary {
-			// special case: "binary" type resolves to []byte in structs, but indicates a streaming response when
-			// specified as the return type of a service, so use "io.ReadCloser".
-			goType = types.IOReadCloserType.GoType(info)
-			imports.AddAll(NewStringSet(types.IOReadCloserType.ImportPaths()...))
```
nmiyake, line 1030: The refactor here doesn't really look like 1:1 -- for example, don't see anything in the newly refactored code that adds the import if needed.

Can you add the functionality in a way that's more obvious without the refactor? I would have expected to see logic that as essentially along the lines of "if the return type if an optional of binary, then special case and return the type that is pointer to IO.ReadCloser"

nmiyake (changes requested):
Thanks for the PR! Think that this is a good start, but a few things:

* Issue should be opened in issue tracker for this and PR should note that it fixes the issue
* In terms of the client, I think the important thing is to handle the returned response code that indicates an empty body and to then return `nil` based on that
* I'm a bit torn on the return type. On the one hand, interfaces support `nil` values, so we could not modify the return type at all for `optional<binary>` and just have the contract be that the function can return nil if the type is `optional<binary>` but not if the return type is `binary`. On the other hand, having the return value be `*io.ReadCloser` does make the "optional" more explicit, and allows users to know whether or not to nil check based just on the return type. For that reason, even though it's a bit less ergonomic from a pure Go standpoint, think that that approach makes sense.
* Ideally, the PR should also add support for generating server code for this as well -- for `optional<binary>` response types, the server handler should be updated accordingly and if the client interface returns `nil`, the server should write a 204 and return nil content. However, if you don't want to add the server logic as part of this PR, would at least open an issue to track this.
* Should add a test case in `integration_test/testgenerated/client` that verifies generated client

conjure/servicewriter.go
nmiyake: Above comment still stands -- don't understand why this PR would modify the "else" clause?

conjure/servicewriter.go
```diff
@@ -1018,25 +1049,20 @@ func returnTypesForEndpoint(endpointDefinition spec.EndpointDefinition, info typ
 	var returnTypes []expression.Type
 	imports := make(StringSet)
 	if endpointDefinition.Returns != nil {
-		var goType string
-		returnBinary, err := isReturnTypeSpecificType(endpointDefinition.Returns, visitors.IsBinary)
+		typer, err := visitors.NewConjureTypeProviderTyper(*endpointDefinition.Returns, info)
 		if err != nil {
 			return nil, nil, err
 		}
+		returnBinary, err := isReturnTypeSpecificType(endpointDefinition.Returns, visitors.IsBinary)
```
nmiyake: Value of "err" should be checked immediately after this call

conjure/servicewriter.go
```diff
@@ -1018,25 +1049,20 @@ func returnTypesForEndpoint(endpointDefinition spec.EndpointDefinition, info typ
...
-			if err != nil {
-				return nil, nil, err
-			}
-			goType = typer.GoType(info)
-			if err != nil {
-				return nil, nil, errors.Wrapf(err, "failed to process return type %q", goType)
-			}
+		}
+		goType := typer.GoType(info)
+		if err != nil {
```
nmiyake: If this is the `err` check that pairs with the `isReturnSpecificType` call, it should be moved up; otherwise, should be removed.

conjure/types/types.go
```diff
@@ -326,3 +326,16 @@ func (f *funcType) ImportPaths() []string {
 	}
 	return importPaths
 }
+
+func MapBinaryTypeToReadCloserType(valType Typer) Typer {
```
nmiyake: Can you add a comment that explains what that this function does?

Also, it doesn't look like there's an effective way to tell if a read closer type was chosen here. Might make sense to update the API to return the `Typer` and any imports that are needed for the typer and then to add the `io` package to the return value if the `IOReadCloserType` or its variant is selected. That way, callers that need the imports can just add the returned imports after calling this.

conjure/servicewriter.go
```diff
@@ -1018,25 +1049,20 @@ func returnTypesForEndpoint(endpointDefinition spec.EndpointDefinition, info typ
...
 		if err != nil {
 			return nil, nil, err
 		}
+		returnBinary, err := isReturnTypeSpecificType(endpointDefinition.Returns, visitors.IsBinary)
 		if returnBinary {
 			// special case: "binary" type resolves to []byte in structs, but indicates a streaming response when
 			// specified as the return type of a service, so use "io.ReadCloser".
-			goType = types.IOReadCloserType.GoType(info)
+			typer = types.MapBinaryTypeToReadCloserType(typer)
 			imports.AddAll(NewStringSet(types.IOReadCloserType.ImportPaths()...))
```
nmiyake: Per feedback on the `MapBinaryTypeToReadCloserType` function, the `IOReadCloserType` import should probably be part of that logic (seems strange to require caller to have to know to add this)

conjure/servicewriter.go
```diff
@@ -811,6 +811,37 @@ func serviceStructMethodBodyAST(endpointDefinition spec.EndpointDefinition, retu
 
 	if returnsBinary {
 		// if endpoint returns binary, return body of response directly
```
nmiyake: Move this comment down to after the `if isOptional` block

conjure/servicewriter.go
```diff
@@ -811,6 +811,37 @@ func serviceStructMethodBodyAST(endpointDefinition spec.EndpointDefinition, retu
...
+		isOptional, err := isReturnTypeSpecificType(endpointDefinition.Returns, visitors.IsOptional)
+		if err != nil {
+			return nil, err
+		}
+		if isOptional {
```
nmiyake, line 818: Would add a comment before this block that explains that if endpoint returns optional of binary, then status code is checked and `nil` is returned if it is `StatusNoContent`

integration_test/testgenerated/client/api/servers.conjure.go
```diff
@@ -146,3 +150,12 @@ func (t *testServiceHandler) HandleBinary(rw http.ResponseWriter, req *http.Requ
...
 	return codecs.Binary.Encode(rw, respArg)
 }
+
+func (t *testServiceHandler) HandleMaybeBinary(rw http.ResponseWriter, req *http.Request) error {
+	respArg, err := t.impl.MaybeBinary(req.Context())
+	if err != nil {
+		return err
+	}
+	rw.Header().Add("Content-Type", codecs.Binary.ContentType())
+	return codecs.Binary.Encode(rw, respArg)
```
AdamDeHovitz: I think this might break here because `Binary.Encode` is expecting an object of type `Reader` not of type `*Reader`. Working on the server code fix.

changelog/@unreleased/pr-191.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: '`optional<binary>` now represented in go as `*ReadCloser` and a received 204 status code is treated as nil'
```
nmiyake: Can you expand this description further? Since these are rolled into the user-facing release notes, would like to make it easier to understand the change based on this content.

For example, something like:

```
Fixes issue where generated Conjure clients did not properly handle empty responses from Conjure endpoints returning `optional<binary>`.

The Conjure specification states that endpoints returning `optional<binary>` can return a 204 status code to indicate an empty optional (to distinguish from the case where the response is a non-empty optional with 0 bytes), but this was not previously supported by the clients generated by conjure-go. This change interprets `optional<binary>` as `*ReadCloser` and returns a `nil` value when the response is a 204.
```

Doesn't have to be this exact wording, but wanted to provide a concrete example of the level of detail that I think would be helpful for developers reading through the release notes to understand the changes.

nmiyake (changes requested):

conjure/servicewriter.go
AdamDeHovitz: I'd prefer to minimize the amount/scope of the special casing here. Part of what allowed this bug to creep in was that the `binary` case was treated handled separately when in reality the binary case is identical to all other cases except we replace references to `[]byte` with `ReadCloser` inside of the `typer` variable.

conjure/servicewriter.go
```diff
@@ -810,6 +810,40 @@ func serviceStructMethodBodyAST(endpointDefinition spec.EndpointDefinition, retu
...
+						expression.Nil,
+						expression.Nil,
+					),
+				},
+			})
+			info.AddImports("net/http")
+			// if endpoint returns binary, return pointer to body of response directly
+			body = append(body, statement.NewReturn(
+				expression.NewSelector(
+					expression.VariableVal("&"+respVar),
```
nmiyake: Sorry, forgot to comment on this in last review: this `body = append(body, statement.NewReturn(...` block is the same as the block outside the "if" statement except for the variable value being `"&"+respVar`. Can we unify that logic and return to simplify branching? Can either create variable before the `if isOptional` and assign accordingly or do the `if isOptional` check after to create the variable val before the `body` assignment.

conjure/servicewriter.go
```diff
@@ -1018,26 +1052,17 @@ func returnTypesForEndpoint(endpointDefinition spec.EndpointDefinition, info typ
...
-				return nil, nil, err
-			}
-			goType = typer.GoType(info)
-			if err != nil {
-				return nil, nil, errors.Wrapf(err, "failed to process return type %q", goType)
-			}
-		}
+
+		// special case: "binary" type resolves to []byte in structs, but indicates a streaming response when
+		// specified as the return type of a service, so replace all nested references with "io.ReadCloser".
```
nmiyake, line 1055: We spoke about this on call and Slack, but forgot the final conclusion: are all nested references replaced (including in lists/sets)? If so, is that correct? If not, comment should be updated because it doesn't describe reality.

nmiyake (changes requested):
Further feedback

conjure/servicewriter.go
AdamDeHovitz: Good catch, I revisited the old code where I generated that `[][]binary` and it now generated `[]ReadCloser` which seems wrong. What I've done is updated this map function to only map `binary` and `optional<binary>` which means that `iterable<binary>` types will be specified as `[][]byte` in the api. I don't think this is correct but it isn't a regression so I'll file an issue to fix it at a later date. I did an internal search and found zero usages of `list<binary>` and only test usages of `set<binary>` which helps explain how this hasn't been an issue thus far.

AdamDeHovitz: Here's that issue: https://github.com/palantir/conjure-go/issues/194

integration_test/testgenerated/client/api/servers.conjure.go
AdamDeHovitz: Once this PR is ready for merge, combining it with this PR should handle this issue: https://github.com/palantir/conjure-go/pull/193

nmiyake (approved):
Thanks, this looks good to me!

nmiyake (changes requested):
Good with this content, but changing to "Request changes" to block merge so that tests can be verified after other PR is merged into this branch

nmiyake (approved):

PR #193, opened 2021-03-16
Improvement: Return 204 status code server side if empty<binary> endpoint returns empty

AdamDeHovitz:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/193)


nmiyake (approved):
Content looks good! Would like tests/generator usage, but believe that is going to be done after this PR is merged into the other branch, so approving this one.

PR #213, opened 2021-07-01
Set test parallelism to 1 to address flake

asanderson15:
## Before this PR
After https://github.com/palantir/conjure-go/pull/200, we've seen a ton of builds flake on the `test` step, specifically on integration_test.go. We haven't figured out why this happens, but it may be a side effect to running go list with that flag in parallel with running that test.

## After this PR
Parallelism is reduced to 1 for the test, which means calls to `go list` no longer overlap with running the integration tests, preventing flakes. 

## Possible downsides?
This roughly doubles the time for this task, but the overall build time should remain the same as it is still dominated by the conjure-verifier task. If this becomes a concern, another option would be to break the integration tests out into their own build task in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/213)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,7 +64,7 @@ jobs:
       - *godel-cache-save
       - *define-tests-dir
       - *mkdir-tests-dir
-      - run: ./godelw test --tags=none --junit-output="$TESTS_DIR/$CIRCLE_PROJECT_REPONAME-tests.xml"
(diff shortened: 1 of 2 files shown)


bmoylan:
@nmiyake for SA, I think it makes sense to merge this but I don't know if you want to do any investigation as to why this would be the case

bmoylan (approved):

nmiyake (approved):
Thanks -- although would be nice to know/address the root cause, don't have any objections with doing this for now.

PR #215, opened 2021-07-09
Improvement: Generate default functions for use with AcceptFuncs union visitor method

jhowarth-pal:
[FieldName]NoopSuccess and ErrorOnUnknown functions are generated for each union type to reduce
boilerplate for default cases.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/215)

### DIFF
--- a/conjure/conjure_test.go
+++ b/conjure/conjure_test.go
@@ -852,6 +852,30 @@ func (u *ExampleUnion) AcceptFuncs(strFunc func(string) error, otherFunc func(st
 	}
 }
 
+func (u *ExampleUnion) StrNoopSuccess(string) error {
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


conjure/conjure_test.go
```diff
@@ -852,6 +852,30 @@ func (u *ExampleUnion) AcceptFuncs(strFunc func(string) error, otherFunc func(st
 	}
 }
 
+func (u *ExampleUnion) StrNoopSuccess(string) error {
+	return nil
+}
+
+func (u *ExampleUnion) OtherNoopSuccess(string) error {
```
jhowarth-pal, line 859: Decided to make these methods rather than static functions to avoid name collisions or requiring the union name within the name

nmiyake (approved):

PR #220, opened 2021-08-17
improvement: Reimplement conjure-go using jennifer and updated type system

bmoylan:
So this ended up being a rewrite of most of the repository, but has the following benefits:
* Introduce a new types abstraction which tracks references between types better than the generated conjure unions.
* Migrate from `goastwriter` to `jennifer`, an open-source code generator with a cleaner syntax and automatic handling of imports
* Fixed some bugs in the type system as a result, especially around `binary` aliases/optionals

I tried to keep the generated code as similar as possible, but couldn't help myself in a couple places:
* Remove need for `_ = resp` in client methods
* Wrap errors with improved messages and status codes

Following this, there are a couple additional bugs in the generated code to fix as their own commits. I also plan on submitting a change to generate reflection-free json methods leveraging the new types.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/220)


nmiyake:
@bmoylan thanks for the PR! Sorry for not reviewing yet despite it being 3 weeks -- I appreciate the PR and how this would improve the code and maintainability of the repository. However, the scale/scope has just made it hard to get started on this. Is this actively blocking something else that you're trying to do, or was it for general maintenance/improvement? Would help to know to get a sense for how to prioritize this. Thanks!

nmiyake (approved):
> Reviewed 138 of 161 files at r1, 32 of 32 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/220)* status: all files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[conjure/snip/imports.go, line 28 at r2](https://reviewable.io/reviews/palantir/conjure-go/220#-MnBrKG9-hNiypWqUgMe:-MnBrKG9-hNiypWqUgMf:be57clc) ([raw file](https://github.com/palantir/conjure-go/blob/bc777d6338928c89c7eec3d1927d682afdf9fcdc/conjure/snip/imports.go#L28)):*
> ```Go
> // A set of imported references included in generated code.
> // Each entry is a func() *jen.Statement, typically the Clone method.
> // This ensures there are no side effects caused by mutating the global variables.
> ```

Thanks for the clarification. It's obviously still possible to completely reassign these (`snip.Context = nil` or `snip.Context = snip.ContextTODO`), but don't think that should be a concern given that the main usage here is local codegen (and don't think it's necessarily worth the extra effort to convert all of these to functions that return the value).

nmiyake (approved):
Thanks for the PR! I looked through all the changes and it appears good to me. Did not look through every line of logic, but things broadly made sense and given the extra context/background we talked through directly I'm comfortable with this change. Thanks for your work on this!

PR #234, opened 2021-11-04
fix: Correctly respond 204 on empty optional values

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/234)

### DIFF
--- a/changelog/@unreleased/pr-234.v2.yml
+++ b/changelog/@unreleased/pr-234.v2.yml
@@ -1,5 +1,10 @@
 type: fix
 fix:
-  description: Correctly respond 204 on empty optional values
+  description: |
+    Generated servers correctly respond 204 for an empty optional response body.
+
+    Server implementations consuming request body of `optional<binary>` (or alias thereof) now use `*io.ReadCloser` for
+    the go type of the body argument. Users must update their interface implementations to the new signature and
+    intentionally handle when the pointer is nil (i.e. the client sent an absent optional value).
   links:
   - https://github.com/palantir/conjure-go/pull/234
--- a/changelog/@unreleased/pr-234.v2.yml
+++ b/changelog/@unreleased/pr-234.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Correctly respond 204 on empty optional values
+  links:
+  - https://github.com/palantir/conjure-go/pull/234
--- a/conjure/serverwriter.go
+++ b/conjure/serverwriter.go
@@ -250,20 +250,39 @@ func astForHandlerMethodDecodeBody(methodBody *jen.Group, argDef *types.Endpoint
 	if argDef == nil {
 		return
 	}
+
+	varName := transforms.SafeName(argDef.Name)
+
+	emptyBodyCondition := jen.Id(reqName).Dot("Body").Op("!=").Nil().Op("&&").
+		Id(reqName).Dot("Body").Op("!=").Add(snip.HTTPNoBody())
+
 	if argDef.Type.IsBinary() {
 		// If the body argument is binary, pass req.Body directly to the impl.
-		methodBody.Id(transforms.SafeName(argDef.Name)).Op(":=").Id(reqName).Dot("Body")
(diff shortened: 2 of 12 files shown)


integration_test/testgenerated/binary/api/servers.conjure.go
```diff
@@ -18,9 +18,10 @@ import (
 type TestService interface {
 	BinaryAlias(ctx context.Context, bodyArg io.ReadCloser) (io.ReadCloser, error)
 	BinaryAliasOptional(ctx context.Context) (*io.ReadCloser, error)
-	BinaryAliasAlias(ctx context.Context, bodyArg io.ReadCloser) (*io.ReadCloser, error)
+	BinaryAliasAlias(ctx context.Context, bodyArg *io.ReadCloser) (*io.ReadCloser, error)
```
bmoylan, line 21: Making the argument a pointer here is a breaking change to code which implements a server resource with optional argument, but I think it is more correct and forces users to contemplate what to do about a nil value.

changelog/@unreleased/pr-234.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: fix
+fix:
+  description: Correctly respond 204 on empty optional values
```
tabboud: Given we're introducing a breaking code change, let's be more descriptive about what changed and what users should do to handle it. We'll most likely want to send out an announcement that this is changing and refer back to these release notes which should be descriptive enough that users won't have to look into the code examples.

conjure/serverwriter.go
```diff
@@ -250,20 +250,39 @@ func astForHandlerMethodDecodeBody(methodBody *jen.Group, argDef *types.Endpoint
 	if argDef == nil {
 		return
 	}
+
+	varName := transforms.SafeName(argDef.Name)
+
+	emptyBodyCondition := jen.Id(reqName).Dot("Body").Op("!=").Nil().Op("&&").
```
tabboud, line 256: Just so I am clear on this piece, if a user provides a nil input argument, it immediately gets mapped to `http.NoBody` and thus we need the `&& req.Body != http.NoBody` so that the argument passed into the impl below is correctly set to nil rather than the `http.NoBody` struct?

conjure/servicewriter.go
```diff
@@ -119,7 +119,11 @@ func astForEndpointParameterArg(argDef *types.EndpointArgumentDefinition, isServ
 		// special case: "binary" types resolve to []byte, but this indicates a streaming parameter when
 		// specified as the request argument of a service, so use "io.ReadCloser".
```
tabboud, line 120: Update comment to indicate we're using `*io.ReadCloser` if it's optional

integration_test/testgenerated/binary/api/servers.conjure.go
tabboud: Agree that this is more correct and forces users to handle the optionality. Similar to other recent changes, we need to be cautious about how this impacts current users. I imagine there are a lot of conjure defs that use optional parameters where it will need to be updated. We should get some more eyes here to raise awareness and provide guidance on the path forward.

tabboud (changes requested):
Overall changes look good to me. Thanks for fixing so quick! 

My only concern is around the breaking change, however it feels justified making this change given the previous behavior was incorrect.

conjure/serverwriter.go
bmoylan: Yep, exactly. I think the reason the stdlib uses NoBody instead of nil is to avoid panics reading from req.Body. I don't think our implementations need to be exposed to that detail, so we use nil.

changelog/@unreleased/pr-234.v2.yml
bmoylan: Updated, let me know what you think!

conjure/servicewriter.go
bmoylan: Done.

tabboud (approved):
Thanks this looks great now!

PR #235, opened 2021-11-04
fix: Respond 204 from endpoints with no return argument

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/235)


tabboud (approved):
👍

PR #240, opened 2021-11-11
fix: Avoid infinite recursion when resolving self-referencing types

bmoylan:
## Before this PR

Defining a type which references itself would cause `resolveType` to panic with stack overflow because it was calling `resolveType` even when it just set the field to the resolved value. This was a noop in all cases except when it caused stack overflow.

## After this PR

Only call `resolveType` if the field is not `unresolvedReferencePlaceholder`

==COMMIT_MSG==
Avoid infinite recursion when resolving self-referencing types.
==COMMIT_MSG==

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/240)

### DIFF
--- a/conjure/types/conjure_definition.go
+++ b/conjure/types/conjure_definition.go
@@ -273,10 +273,8 @@ func (t *namedTypes) resolveType(typeI Type) error {
 			} else {
 				return errors.Errorf("Unresolved optional type reference %s %s", unresolved.Ref.Package, unresolved.Ref.Name)
 			}
-		} else {
-			if err := t.resolveType(v.Item); err != nil {
-				return err
-			}
+		} else if err := t.resolveType(v.Item); err != nil {
+			return err
 		}
 	case *List:
 		if unresolved, ok := v.Item.(unresolvedReferencePlaceholder); ok {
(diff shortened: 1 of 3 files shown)


conjure/types/conjure_definition.go
```diff
@@ -273,17 +273,23 @@ func (t *namedTypes) resolveType(typeI Type) error {
 			} else {
 				return errors.Errorf("Unresolved optional type reference %s %s", unresolved.Ref.Package, unresolved.Ref.Name)
 			}
+		} else {
```
nmiyake: Somewhat of a nit, but think `else if` construct may be better in these cases to make intended control flow clearer

nmiyake (approved):

conjure/types/conjure_definition.go
```diff
@@ -493,7 +500,3 @@ func (unresolvedReferencePlaceholder) Code() *jen.Statement {
 func (unresolvedReferencePlaceholder) String() string {
 	panic("unresolvedReferencePlaceholder does not implement methods")
 }
-
-func (unresolvedReferencePlaceholder) Equals(t Type) bool {
```
bmoylan, line 497: this was just not needed (Equals got removed from the type interface)

PR #242, opened 2021-11-12
fix: Update jennifer to fix conjure-docs struct tags

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/242)


vendor/modules.txt
```diff
@@ -1,4 +1,4 @@
-# github.com/dave/jennifer v1.4.1
+# github.com/dave/jennifer v1.4.2-0.20211112003305-45cc0b7eb71a
```
bmoylan, line 1: Current commit on develop, which includes https://github.com/dave/jennifer/pull/83 (I don't think Dave is tagging releases anymore)

nmiyake (approved):

PR #243, opened 2021-11-15
Fix: fix IsUnknown for enums with no declared values

tabboud:
## Before this PR 
Enums with an empty list of values produced non-compiling code as seen below.
```go
// IsUnknown returns false for all known variants of EmptySliceDeclFix and true otherwise.
func (e EmptySliceDeclFix) IsUnknown() bool {
	switch e.val {
	case :
		return false
	}
	return true
}
```

Prior to the migrating to Jennifer the code was as follows:
```go
// IsUnknown returns false for all known variants of EmptySliceDeclFix and true otherwise.
func (e EmptySliceDeclFix) IsUnknown() bool {
	switch e.val {
	default:
		return false
	}
	return true
}
```

## After this PR

==COMMIT_MSG==
Fix the IsUnknown method for enums with no declared values
==COMMIT_MSG==

**UPDATE** 

This change now just fixes the compile breaks and retains the previous behavior

---

~~This fixes the code gen, but also changes the behavior slightly to what I would consider _more correct_. As seen above, the previously generated code would always return `false` for the `IsUnknown` method. However, given that there are no known values for this enum, it seems more correct to always return `true` given there are no known variants that are allowed.~~

## Possible downsides?
- ~~Behavior change for the `IsUnknown` function which may be dependent upon by consumers.~~

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/243)

### DIFF
--- a/conjure/enumwriter.go
+++ b/conjure/enumwriter.go
@@ -92,8 +92,9 @@ func astForEnumIsUnknown(typeName string, values []*types.Field) *jen.Statement 
 					}
 				}).Block(jen.Return(jen.False()))
 			}
-		methodBody.Return(jen.True())
-		})})
+			methodBody.Return(jen.True())
+		})
+	})
 }
 
 func astForEnumValueMethod(typeName string) *jen.Statement {
--- a/conjure/enumwriter.go
+++ b/conjure/enumwriter.go
@@ -81,17 +81,17 @@ func astForEnumIsUnknown(typeName string, values []*types.Field) *jen.Statement 
 	return jen.Commentf("IsUnknown returns false for all known variants of %s and true otherwise.", typeName).
 		Line().
 		Func().
-		Params(jen.Id(enumReceiverName).Id(typeName)).Id("IsUnknown").Params().Params(jen.Bool()).Block(
-		jen.Switch(jen.Id(enumReceiverName).Dot(enumStructFieldName)).Block(
-			jen.CaseFunc(func(conds *jen.Group) {
-				for _, valDef := range values {
-					conds.Id(typeName + "_" + valDef.Name)
-				}
-			}).
-				Block(jen.Return(jen.False())),
-		),
-		jen.Return(jen.True()),
-	)
+		Params(jen.Id(enumReceiverName).Id(typeName)).Id("IsUnknown").Params().Params(jen.Bool()).BlockFunc(func(methodBody *jen.Group) {
+		if len(values) > 0 {
+			methodBody.Switch(jen.Id(enumReceiverName).Dot(enumStructFieldName)).Block(
+				jen.CaseFunc(func(conds *jen.Group) {
+					for _, valDef := range values {
+						conds.Id(typeName + "_" + valDef.Name)
+					}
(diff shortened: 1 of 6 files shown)


bmoylan (approved):
Thanks for fixing!

> Reviewed 5 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/243#-:-Mo_G-1q2OxN2nj1setp:b-65fm7p)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nmiyake)

nmiyake:
Is it possible to separate these changes so that we fix the compile error in one commit and change the behavior in another? I'm a bit concerned about completely inverting the behavior if there are any consumers, so would like to at least do a usage search before proceeding with the behavior change.

integration_test/testgenerated/objects/objects_test.go
```diff
@@ -409,3 +409,9 @@ func TestEnumUnknownValue(t *testing.T) {
 func TestEnumValues(t *testing.T) {
 	assert.Equal(t, []api.Enum_Value{api.Enum_VALUE, api.Enum_VALUES, api.Enum_VALUES_1, api.Enum_VALUES_1_1, api.Enum_VALUE1, api.Enum_VALUE2}, api.Enum_Values())
 }
+
+func TestEmptyValuesEnumIsAlwaysUnknown(t *testing.T) {
+	assert.False(t, api.New_EmptyValuesEnum("test-value").IsUnknown())
+	// TODO(tabboud): Explicit unknown values should always return true
+	assert.False(t, api.New_EmptyValuesEnum(api.EmptyValuesEnum_UNKNOWN).IsUnknown())
```
tabboud, line 416: This showcases the previous behavior which is entirely incorrect, but will be fixed in a followup PR

conjure/enumwriter.go
```diff
@@ -81,17 +81,19 @@ func astForEnumIsUnknown(typeName string, values []*types.Field) *jen.Statement
...
-				}
-			}).
-				Block(jen.Return(jen.False())),
-		),
-		jen.Return(jen.True()),
-	)
+		Params(jen.Id(enumReceiverName).Id(typeName)).Id("IsUnknown").Params().Params(jen.Bool()).BlockFunc(func(methodBody *jen.Group) {
+		methodBody.Switch(jen.Id(enumReceiverName).Dot(enumStructFieldName)).BlockFunc(func(switchBlock *jen.Group) {
+			if len(values) == 0 {
+				switchBlock.Default().Return(jen.False())
```
tabboud, line 87: This matches the previous behavior, by injecting a default false block

bmoylan (approved):
> Reviewed 3 of 4 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/243)* status: all files reviewed, 2 unresolved discussions (waiting on @nmiyake)

nmiyake (approved):

PR #245, opened 2021-11-16
Fix: fix IsUnknown behavior to only return false for known variants

tabboud:
## Before this PR 
Enums with no values would always return `false` for `IsUnknown`, however by definition there are no known variants of the enum if none are listed, thus `IsUnknown` should always return true.

Given the following conjure enum with no values
```yaml
types:
  definitions:
    objects:
      EmptySliceDeclFix:
        values: []
```

The current generated code is as follows:

```go
// IsUnknown returns false for all known variants of EmptySliceDeclFix and true otherwise.
func (e EmptySliceDeclFix) IsUnknown() bool {
	switch e.val {
	default:
		return false
	}
	return true
}
```

After this change the code becomes:
```go
// IsUnknown returns false for all known variants of EmptySliceDeclFix and true otherwise.
func (e EmptySliceDeclFix) IsUnknown() bool {
	return true
}
```

## After this PR

==COMMIT_MSG==
Fix the IsUnknown generated code behavior to only return false for known variants
==COMMIT_MSG==

This fixes the behavior slightly to what I would consider _more correct_. As seen above, the previously generated code would always return `false` for the `IsUnknown` method. However, given that there are no known values for this enum, it seems more correct to always return `true` given there are no known variants that are allowed.

## Possible downsides?
- Behavior change for the `IsUnknown` function which may be dependent upon by consumers, although the previous behavior was incorrect.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/245)

### DIFF
--- a/changelog/@unreleased/pr-245.v2.yml
+++ b/changelog/@unreleased/pr-245.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Fix the IsUnknown generated code behavior to only return false for
+    known variants
+  links:
+  - https://github.com/palantir/conjure-go/pull/245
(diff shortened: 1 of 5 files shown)


bmoylan (approved):
> Reviewed 4 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/245)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nmiyake)

PR #246, opened 2021-11-20
fix: Fix client/server methods for request body of alias<optional<T>>

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/246)

### DIFF
--- a/conjure/serverwriter.go
+++ b/conjure/serverwriter.go
@@ -443,7 +443,7 @@ func astForHandlerExecImplAndReturn(g *jen.Group, serviceName string, endpointDe
 		}
 
 		// Empty optionals return a 204 (No Content) response
-		g.If(respVal.Clone().Op("==").Nil()).Block(
+		g.If(respVal.Op("==").Nil()).Block(
 			jen.Id(responseWriterVarName).Dot("WriteHeader").Call(snip.HTTPStatusNoContent()),
 			jen.Return(jen.Nil()),
 		)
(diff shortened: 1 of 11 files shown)


conjure/serverwriter.go
```diff
@@ -433,12 +433,20 @@ func astForHandlerExecImplAndReturn(g *jen.Group, serviceName string, endpointDe
...
+			// If the response type is named (i.e. an alias), check the inner Value field for absence.
+			respVal = respVal.Dot("Value")
+		} else {
+			// If the response is not named, it's a pointer to the underlying type. Dereference it for the Encoder.
+			respArg = jen.Op("*").Add(respArg.Clone())
+		}
+
 		// Empty optionals return a 204 (No Content) response
-		g.If(respArg.Clone().Op("==").Nil()).Block(
+		g.If(respVal.Clone().Op("==").Nil()).Block(
```
tabboud: Do we need this additional `.Clone()` since we now clone the `respArg` on line 436?

bmoylan: I guess we don't. Removed!

integration_test/testgenerated/server/server-service.yml
```diff
@@ -34,6 +34,11 @@ services:
         args:
           body: optional<CustomObject>
         returns: optional<CustomObject>
+      echoOptionalAlias:
```
tabboud, line 39: Can you add a test case for an option alias that contains a collection in addition to this primitive type? I believe this will introduce a compile time break since we check `returnVal == nil` in  `servicewriter.go` but the returnVal will be the optional type i.e. a struct which can't check for nil. So we may need to introspect on the `.Value` field there as well

bmoylan: You are right! I've added a fix and test. If something is optional we don't want to check for nil at all (204s skip unmarshaling because there is no body) so I think it's most correct to remove it (rather than inspect the internal field at all)

tabboud (approved):
LGTM

PR #248, opened 2021-11-22
Fix: Fix code generation of request param parsing for external types

### LINKED ISSUE #247: Break in code-gen for external types used in path/query params
## What happened?
A panic was encountered after upgrading from `v6.5.0` to `v.6.6.0`, which includes the code-gen change to jennifer from goastwriter.
Of note, this appears when an API uses an external type in either path or query parameters.

```
panic: unrecognized type BackupId (string)

goroutine 1 [running]:
github.com/palantir/conjure-go/v6/conjure.astForDecodeHTTPParamInternal(0xc00038fe60, {0xc00023a588, 0x8}, {0x1465920, 0xc00041e080}, {0xc00023a588, 0x8}, {0x145a6f0, 0xc000389a58}, 0x0)
        /Volumes/git/go/src/github.com/palantir/conjure-go/conjure/serverwriter.go:387 +0x980d
github.com/palantir/conjure-go/v6/conjure.astForDecodeHTTPParam(...)
        /Volumes/git/go/src/github.com/palantir/conjure-go/conjure/serverwriter.go:289
github.com/palantir/conjure-go/v6/conjure.astForHandlerMethodQueryParam(0x0, 0xc000028d80)
        /Volumes/git/go/src/github.com/palantir/conjure-go/conjure/serverwriter.go:246 +0x1247
github.com/palantir/conjure-go/v6/conjure.astForHandlerMethodQueryParams(0x0, {0xc00034e378, 0x1, 0x0})
        /Volumes/git/go/src/github.com/palantir/conjure-go/conjure/serverwriter.go:234 +0x38
github.com/palantir/conjure-go/v6/conjure.astForHandlerMethodBody(0x100e367, {0xc0004e9020, 0xb}, 0xc0003a4bb0)
        /Volumes/git/go/src/github.com/palantir/conjure-go/conjure/serverwriter.go:149 +0x115
github.com/palantir/conjure-go/v6/conjure.astForHandlerMethods.func1(0xc00005ad80)
        /Volumes/git/go/src/github.com/palantir/conjure-go/conjure/serverwriter.go:137 +0x2c
github.com/dave/jennifer/jen.(*Statement).BlockFunc(0xc0002beee8, 0xc0005cb6a8)
        /Volumes/git/go/src/github.com/palantir/conjure-go/vendor/github.com/dave/jennifer/jen/generated.go:234 +0x89
github.com/palantir/conjure-go/v6/conjure.astForHandlerMethods(0xc00033e1e0)
        /Volumes/git/go/src/github.com/palantir/conjure-go/conjure/serverwriter.go:136 +0x1265
github.com/palantir/conjure-go/v6/conjure.writeServerType(0xc000285980, 0xc00033e1e0)
        /Volumes/git/go/src/github.com/palantir/conjure-go/conjure/serverwriter.go:54 +0x597
github.com/palantir/conjure-go/v6/conjure.GenerateOutputFiles({0x1, {0x1edea50, 0x0, 0x0}, {0xc00019a540, 0x4, 0x4}, {0xc0001f4400, 0x1, 0x4}, ...}, ...)
        /Volumes/git/go/src/github.com/palantir/conjure-go/conjure/conjure.go:95 +0x1005
github.com/palantir/conjure-go/v6/conjure.Generate({0x1, {0x1edea50, 0x0, 0x0}, {0xc00019a540, 0x4, 0x4}, {0xc0001f4400, 0x1, 0x4}, ...}, ...)
        /Volumes/git/go/src/github.com/palantir/conjure-go/conjure/conjure.go:29 +0x65
main.run({0x13d4413, 0x480}, {0x13b2b14, 0x6})
        /Volumes/git/go/src/github.com/palantir/conjure-go/integration_test/testgenerated/generate.go:61 +0x14f
main.main()
        /Volumes/git/go/src/github.com/palantir/conjure-go/integration_test/testgenerated/generate.go:33 +0xa5
exit status 2
```

Code should be generated as it was prior to `v6.6.0`

tabboud:
## Before this PR
Break in code-gen for APIs with external types.

## After this PR

==COMMIT_MSG==
Fix code generation of request param parsing for external types.
Fixes: #247 
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/248)

### DIFF
--- a/changelog/@unreleased/pr-248.v2.yml
+++ b/changelog/@unreleased/pr-248.v2.yml
@@ -1,0 +1,7 @@
+type: fix
+fix:
+  description: |-
+    Fix code generation of request param parsing for external types.
+    Fixes: #247
+  links:
+  - https://github.com/palantir/conjure-go/pull/248
--- a/conjure/serverwriter.go
+++ b/conjure/serverwriter.go
@@ -383,6 +383,8 @@ func astForDecodeHTTPParamInternal(methodBody *jen.Group, argName string, argTyp
 		)
 	case *types.Map, *types.ObjectType, *types.UnionType:
 		panic(fmt.Sprintf("unsupported complex type for http param %v", argType))
+	case *types.External:
+		astForDecodeHTTPParamInternal(methodBody, argName, typVal.Fallback, outVarName, inStrExpr, depth+1)
 	default:
 		panic(fmt.Sprintf("unrecognized type %v", argType))
 	}
--- a/conjure/serverwriter_test.go
+++ b/conjure/serverwriter_test.go
@@ -19,6 +19,7 @@ import (
 	"testing"
 
 	"github.com/dave/jennifer/jen"
+	"github.com/palantir/conjure-go/v6/conjure-api/conjure/spec"
 	"github.com/palantir/conjure-go/v6/conjure/types"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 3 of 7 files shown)


bmoylan:
___
*[conjure/serverwriter.go, line 387 at r2](https://reviewable.io/reviews/palantir/conjure-go/248#-Mp8y3rg3AZ-J4LHbZZC:-Mp8y3rg3AZ-J4LHbZZD:byem149) ([raw file](https://github.com/palantir/conjure-go/blob/b4313555607a6c9909a5572e963ae465c38f0fac/conjure/serverwriter.go#L387)):*
> ```Go
> 		panic(fmt.Sprintf("unsupported complex type for http param %v", argType))
> 	case *types.External:
> 		astForDecodeHTTPParamInternal(methodBody, argName, typVal.Fallback, outVarName, inStrExpr, depth+1)
> ```

do we always want to use the fallback? Should we try to actually render in the external type?

tabboud (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/248)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[conjure/serverwriter.go, line 387 at r2](https://reviewable.io/reviews/palantir/conjure-go/248#-Mp8y3rg3AZ-J4LHbZZC:-Mp91RJX7e-bhJrYO9gh:b-60aklz) ([raw file](https://github.com/palantir/conjure-go/blob/b4313555607a6c9909a5572e963ae465c38f0fac/conjure/serverwriter.go#L387)):*

do we always want to use the fallback? Should we try to actually render in the external type?
</blockquote></details>

That's a good question and one I was thinking about while adding this. The previous logic didn't seem to follow the conjure spec or at least made other assumptions (specifically splitting on `:`) when inspecting the defined type and then would default to the fallback.
splitting on `:`: https://github.com/palantir/conjure-go/blob/v6.5.0/conjure/types/types.go#L254
defaulting to fallback: https://github.com/palantir/conjure-go/blob/v6.5.0/conjure/visitors/conjuretype_visit_external.go#L32-L35

Since I couldn't find the reason why we split on `:` I ended up leaving it off. Do you happen to know why or how this was used previously?

bmoylan:
___
*[conjure/serverwriter.go, line 387 at r2](https://reviewable.io/reviews/palantir/conjure-go/248#-Mp8y3rg3AZ-J4LHbZZC:-MpCYMzo9c5QbbU0UsKw:bpto0fn) ([raw file](https://github.com/palantir/conjure-go/blob/b4313555607a6c9909a5572e963ae465c38f0fac/conjure/serverwriter.go#L387)):*
<details><summary><i>Previously, tabboud (Tony Abboud) wrote…</i></summary><blockquote>

That's a good question and one I was thinking about while adding this. The previous logic didn't seem to follow the conjure spec or at least made other assumptions (specifically splitting on `:`) when inspecting the defined type and then would default to the fallback.
splitting on `:`: https://github.com/palantir/conjure-go/blob/v6.5.0/conjure/types/types.go#L254
defaulting to fallback: https://github.com/palantir/conjure-go/blob/v6.5.0/conjure/visitors/conjuretype_visit_external.go#L32-L35

Since I couldn't find the reason why we split on `:` I ended up leaving it off. Do you happen to know why or how this was used previously?
</blockquote></details>

Yeah I tried to preserve the existing logic but don't actually know of particular use cases of this. I think it exists because we don't have a good mapping from conjure package name to go import path so it allows for specifying a non-conjure-standard string we can turn into a go type.

bmoylan (approved):

tabboud:
Brad and I talked offline and agreed that we should move forward with the approach of always using the fallback rather than try to preserve the external type.
This decision was based on the following

- The conjure-spec strongly discourages the use of external types: https://palantir.github.io/conjure/#/docs/spec/conjure_definitions?id=externaltypedefinition
- There is a fairly strict type set for HTTP params which means we would end up using the fallback anyways: https://palantir.github.io/conjure/#/docs/spec/conjure_definitions?id=argumentdefinition
- There is no native support for Go external types. We have logic that is non-spec compliant to infer a package name and type from the spec name: https://github.com/palantir/conjure-go/blob/7bcaee85b3b6ecdea93a3ef3b14daf5cda6a744b/conjure/types/types.go#L305-L312

There are some options we have to make this work with the declared type, however we will have to work around the spec or add some more hacky code paths to make things work. I have also found 0 usages of the Go external types, so the old code path (pre-jennifer code gen) would use the fallback the majority of the time. Given all of this we are moving forward with the current approach.

PR #250, opened 2021-11-29
fix: Correctly handle binary return type with auth enabled

okushchenko:
## Before this PR
See updated [](integration_test/testgenerated/auth/auth-service.yml) file for an example of a Conjure definition that results in generated code that fails to compile. The error looks like this:

```
services.conjure.go:1:1: cannot use defaultReturnVal (type []byte) as type io.ReadCloser in return argument:
	[]byte does not implement io.ReadCloser (missing Close method)
```

The bug is only reproducible for Conjure endpoints that return a `binary` or `optional<binary>` type and have auth enabled.

I think it was introduced in https://github.com/palantir/conjure-go/releases/tag/v6.6.0 or later, as previous version didn't have this bug.

## After this PR
==COMMIT_MSG==
Correctly handle binary return type with auth enabled.
==COMMIT_MSG==

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/250)

### DIFF
--- a/changelog/@unreleased/pr-250.v2.yml
+++ b/changelog/@unreleased/pr-250.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Correctly handle binary return type with auth enabled.
+  links:
+  - https://github.com/palantir/conjure-go/pull/250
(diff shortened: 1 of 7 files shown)


tabboud (approved):
Thanks for the fix and separating the commits by fix and refactor which helps a ton when reviewing!

PR #252, opened 2021-11-30
Fix: Generate alias encoding methods based on the external fallback type

### LINKED ISSUE #251: Alias of an external type with an any fallback produces non-compiling code
## What happened?

An alias definition for an external type that uses `any` as the fallback will generate serde methods for the type. Since the type is `any`, which maps to an `interface{}`, methods cannot be added to the type and thus results in a compile time error.

Here is a short repro of the issue:

Conjure definition
```yml
types:
  imports:
    ExternalLong:
      base-type: any
      external:
        java: java.lang.Long
  definitions:
    default-package: api
    objects:
      LongAlias:
        alias: ExternalLong
```

When compiled with conjure-go >= `v6.6.0` the following will be produced:
```go
// This file was generated by Conjure and should not be manually edited.

package api

import (
	safejson "github.com/palantir/pkg/safejson"
	safeyaml "github.com/palantir/pkg/safeyaml"
)

type LongAlias interface{}

func (a LongAlias) MarshalJSON() ([]byte, error) {
	return safejson.Marshal(interface{}(a))
}

func (a *LongAlias) UnmarshalJSON(data []byte) error {
	var rawLongAlias interface{}
	if err := safejson.Unmarshal(data, &rawLongAlias); err != nil {
		return err
	}
	*a = LongAlias(rawLongAlias)
	return nil
}

func (a LongAlias) MarshalYAML() (interface{}, error) {
	jsonBytes, err := safejson.Marshal(a)
	if err != nil {
		return nil, err
	}
	return safeyaml.JSONtoYAMLMapSlice(jsonBytes)
}

func (a *LongAlias) UnmarshalYAML(unmarshal func(interface{}) error) error {
	jsonBytes, err := safeyaml.UnmarshalerToJSONBytes(unmarshal)
	if err != nil {
		return err
	}
	return safejson.Unmarshal(jsonBytes, *&a)
}
```

Since the fallback type is an interface, no methods should be serde methods should be generated, so we would result in just the type definition.
```go
// This file was generated by Conjure and should not be manually edited.

package api

type LongAlias interface{}
```

The behavior above matches what was generated prior to conjure-go `v6.6.0`

tabboud:
## Before this PR 
See #251 for more information

## After this PR

==COMMIT_MSG==
Generate alias encoding methods based on the external fallback type
Fixes #251
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/252)

### DIFF
--- a/changelog/@unreleased/pr-252.v2.yml
+++ b/changelog/@unreleased/pr-252.v2.yml
@@ -1,0 +1,7 @@
+type: fix
+fix:
+  description: |-
+    Generate alias encoding methods based on the external fallback type
+    Fixes #251
+  links:
+  - https://github.com/palantir/conjure-go/pull/252
--- a/conjure/aliaswriter.go
+++ b/conjure/aliaswriter.go
@@ -113,6 +113,8 @@ func isSimpleAliasType(t types.Type) bool {
 		return isSimpleAliasType(v.Item)
 	case *types.AliasType:
 		return isSimpleAliasType(v.Item)
+	case *types.External:
+		return isSimpleAliasType(v.Fallback)
 	default:
 		return false
 	}
--- a/integration_test/testgenerated/objects/api/aliases.conjure.go
+++ b/integration_test/testgenerated/objects/api/aliases.conjure.go
@@ -46,8 +46,42 @@ func (a *BinaryAlias) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return safejson.Unmarshal(jsonBytes, *&a)
 }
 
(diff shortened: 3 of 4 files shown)


bmoylan (approved):
> Reviewed 3 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/252)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tabboud)

PR #255, opened 2021-12-10
fix: Marshal JSON 'null' for empty optionals

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/255)


integration_test/testgenerated/objects/objects_test.go
```diff
@@ -112,6 +112,7 @@ func TestMarshal(t *testing.T) {
 		{api.Compound{}, `{"obj":{"mapVar":{},"listVar":[],"multiDim":[]}}`, "obj:\n  mapVar: {}\n  listVar: []\n  multiDim: []\n"},
 		{api.BooleanIntegerMap{}, `{"map":{}}`, "map: {}\n"},
 		{api.BooleanIntegerMap{Map: map[boolean.Boolean]int{false: 1, true: 2}}, `{"map":{"false":1,"true":2}}`, "map:\n  \"false\": 1\n  \"true\": 2\n"},
+		{api.OptionalStructAlias{}, "null", "null\n"},
```
bmoylan, line 115: Previously this fails with
```
json: error calling MarshalJSON for type api.OptionalStructAlias: unexpected end of JSON input
```

tabboud (approved):
I believe this is the link to the spec where this is referenced: https://palantir.github.io/conjure/#/docs/spec/wire?id=_52-container-types

The change looks good to me, but we may want to be on the lookout for any users depending on the current non compliant behavior.

PR #256, opened 2021-12-10
fix: Escape server handler variable names with Arg

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/256)


conjure/serverwriter.go
```diff
@@ -190,10 +190,10 @@ func astForHandlerMethodPathParams(methodBody *jen.Group, pathParams []*types.En
 }
 
 func astForHandlerMethodPathParam(methodBody *jen.Group, argDef *types.EndpointArgumentDefinition) {
-	strVar := transforms.SafeName(argDef.ParamID) + "Str"
```
tabboud, line 193: We can remove the `SafeName` function altogether now since it's not used anymore and since we're appending the `Arg` we shouldn't have collisions.

tabboud (approved):

PR #258, opened 2021-12-11
fix: Decode alias-typed HTTP params recursively based on their underlying type

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/258)

### DIFF
--- a/conjure/serverwriter.go
+++ b/conjure/serverwriter.go
@@ -220,10 +220,9 @@ func astForHandlerMethodHeaderParams(methodBody *jen.Group, headerParams []*type
 
 func astForHandlerMethodHeaderParam(methodBody *jen.Group, argDef *types.EndpointArgumentDefinition) {
 	var queryVar jen.Code
-	switch argDef.Type.(type) {
-	case *types.List:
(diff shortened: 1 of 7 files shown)


conjure/serverwriter.go
```diff
@@ -354,21 +352,28 @@ func astForDecodeHTTPParamInternal(methodBody *jen.Group, argName string, argTyp
 				g.Id(outVarName).Op("=").Append(jen.Id(outVarName), jen.Id("convertedVal"))
 			})
 		}
+	case *types.Set:
```
bmoylan, line 355: same as types.List case

tabboud, line 355: Can you add a unit test for sets in `serverwriter_test.go` since it looks like we don't have any yet.

tabboud (commented):
Change looks good, just 1 request for another test case

conjure/serverwriter.go
bmoylan: Yup, done!

tabboud (approved):

PR #261, opened 2022-01-07
improvement: Add omitempty to optionals' json struct tag

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/261)


nmiyake (commented):
Is there a particular motivation here? I'm not opposed to this, but in some instances having fields explicitly blank can help to understand the fields that exist as well. Just wanted to understand if this is to match previous behavior (or behavior in other implementations), to reduce size, for aesthetics purposes, in order to provide compatibility with something else, etc.

PR #297, opened 2022-04-18
Improvement: Add support for CLI generation

asanderson15:
==COMMIT_MSG==
Add support for CLI generation
==COMMIT_MSG==

* Generate CLI bindings for Cobra CLI, enabling easy CLI support for any conjured endpoint

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/297)


bmoylan (changes requested):
> Reviewed 37 of 49 files at r1, 1 of 2 files at r2, 3 of 12 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/297)* status: 41 of 50 files reviewed, 9 unresolved discussions (waiting on @asanderson15 and @bmoylan)

___
*[`go.mod` line 37 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0poVouDIb5-g2YF17J:-N0poVouDIb5-g2YF17K:b-meqfzi) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/go.mod#L37)):*
> ```XML
> )
> 
> require github.com/stretchr/objx v0.1.1 // indirect
> ```

I'm surprised go mod left this here, but should it be in the block of indirect deps below?
___
*[`conjure/cliwriter.go` line 114 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0pqkGl7HOT5GBkzPtN:-N0pqkGl7HOT5GBkzPtO:bnsbcl2) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/conjure/cliwriter.go#L114)):*
> ```Go
> }
> 
> // astForGetCLIContextBody implements the getCLIContext function
> ```

What would you think about supporting a `-q/--quiet` which doesn't do any logging, or maybe logging to standard error instead of standard out? I think this tool could be used in bash pipelines where the stdout logging mucks up that stream.
___
*[`conjure/cliwriter.go` line 233 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0ps91gB1TITOp3CK0e:-N0ps91gB1TITOp3CK0f:b548j0e) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/conjure/cliwriter.go#L233)):*
> ```Go
> 		Dot("PersistentFlags").Call().
> 		Dot("String").Call(
> 		jen.Lit(confFlagName), jen.Lit(defaultConfigFilePath), jen.Lit("The configuration file is optional. The default path is ./var/conf/configuration.yml.")).Line()
> ```

This may not be worth the effort, but we could consider using [viper](https://github.com/spf13/viper) rather than pflag because it supports arbitrarily deep overrides on nested structures. I haven't used it, but I imagine it would allow doing something like `--client.uris https://localhost:443 --client.api-token abcd1234` 
___
*[`conjure/cliwriter.go` line 379 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0pthS38lEzijLaJn7q:-N0pthS38lEzijLaJn7r:b-95qhoi) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/conjure/cliwriter.go#L379)):*
> ```Go
> 
> 	// TODO: Add support for reading file from path as binary input.
> 	// Note that this code path should not be hit because we skip generating code for endpoints with binary params
> ```

I think "idiomatic" here (probably explained in the arg's docstring) would be to accept either base64 in the flag value or `@/path/to/file.dat` to indicate it should read a file
___
*[`conjure/cliwriter.go` line 381 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0pu1pA9fzU-x3UOi1U:-N0pu1pA9fzU-x3UOi1V:bqt1xm3) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/conjure/cliwriter.go#L381)):*
> ```Go
> 	// Note that this code path should not be hit because we skip generating code for endpoints with binary params
> 	if param.Type.IsBinary() {
> 		file.Id("panic").Call(jen.Lit("Commands with binary arguments are not yet supported."))
> ```

Can we return an error instead of panicking? Stacktraces in the terminal scare people
___
*[`conjure/cliwriter.go` line 438 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0pusW01XfDqH5SiLLX:-N0pusW01XfDqH5SiLLY:b-z802s1) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/conjure/cliwriter.go#L438)):*
> ```Go
> 	file.Id(argBytesName).Op(":=").Index().Byte().Parens(flagVar)
> 	file.If(
> 		jen.Err().Op(":=").Add(snip.CGRCodecsJSON().Dot("Decode")).Call(
> ```

There's a `Unmarshal` method so you don't have to create a reader (unless this is future-proofing for reading from files...)
___
*[`conjure/snip/imports.go` line 62 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0poxRH7x74SJv39xLc:-N0poxRH7x74SJv39xLd:blpn14p) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/conjure/snip/imports.go#L62)):*
> ```Go
> // This ensures there are no side effects caused by mutating the global variables.
> var (
> 	ByteReader          = jen.Qual("bytes", "NewReader")
> ```

needs `.Clone`
___
*[`conjure/snip/imports.go` line 78 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0pp4Sc7IBit42FX2le:-N0pp4Sc7IBit42FX2lf:b9ke30o) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/conjure/snip/imports.go#L78)):*
> ```Go
> 	IOCopy              = jen.Qual("io", "Copy").Clone()
> 	IOUtilReadAll       = jen.Qual("io/ioutil", "ReadAll").Clone
> 	IOUtilReadFile      = jen.Qual("io/ioutil", "ReadFile").Clone
> ```

ioutil is deprecated, let's use `os.ReadFile`
___
*[`integration_test/testgenerated/cli/cli-service.yml` line 16 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0ppxFo3gP5orO00UHH:-N0ppxFp5ripwSZJ1Z89:b-r5a9co) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/integration_test/testgenerated/cli/cli-service.yml#L16)):*
> ```YAML
>       base-type: integer
>       external:
>         java: com.palantir.rescue.api.ExternalInteger
> ```

nit: let's keep `rescue` out of this repo

asanderson15 (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/297)* status: 41 of 50 files reviewed, 8 unresolved discussions (waiting on @bmoylan)

___
*[`go.mod` line 37 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0poVouDIb5-g2YF17J:-N0q8jqX1cZJ5YzbfF2M:b-60nahp) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/go.mod#L37)):*

I'm surprised go mod left this here, but should it be in the block of indirect deps below?
</blockquote></details>

weird, will fix
___
*[`conjure/cliwriter.go` line 114 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0pqkGl7HOT5GBkzPtN:-N0q8MLm7Qo9t0C5R6RM:b-lpr6wu) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/conjure/cliwriter.go#L114)):*

What would you think about supporting a `-q/--quiet` which doesn't do any logging, or maybe logging to standard error instead of standard out? I think this tool could be used in bash pipelines where the stdout logging mucks up that stream.
</blockquote></details>

I'm going to track this as an FR (already have a decent list of small improvements like this).

I think I'd go the other direction and add a `-v/--verbose` flag and make it quiet by default. The main logging tends to be tracelog, which are useful for debugging, but you don't really need it by default.
___
*[`conjure/cliwriter.go` line 233 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0ps91gB1TITOp3CK0e:-N0q9Lsj8fYo83e15qdk:bqwiivl) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/conjure/cliwriter.go#L233)):*

This may not be worth the effort, but we could consider using [viper](https://github.com/spf13/viper) rather than pflag because it supports arbitrarily deep overrides on nested structures. I haven't used it, but I imagine it would allow doing something like `--client.uris https://localhost:443 --client.api-token abcd1234` 
</blockquote></details>

Yeah, I think it's worth considering in the future. From poking around, it seems like it would require a bit deeper integration between flags and arg deserialization though, which may or may not just work - would need to test it out more. 
___
*[`conjure/cliwriter.go` line 379 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0pthS38lEzijLaJn7q:-N0qAlmA7cEVvx_1dHDR:b-c2eeuu) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/conjure/cliwriter.go#L379)):*

I think "idiomatic" here (probably explained in the arg's docstring) would be to accept either base64 in the flag value or `@/path/to/file.dat` to indicate it should read a file
</blockquote></details>

Yep, exactly - I plan to add support for this as a follow-on. I also want to support this for collections and objects to make it easier to pipe output to file, edit, and submit an update.
___
*[`conjure/cliwriter.go` line 381 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0pu1pA9fzU-x3UOi1U:-N0qA-X28xBzbenWQOhC:breov7y) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/conjure/cliwriter.go#L381)):*

Can we return an error instead of panicking? Stacktraces in the terminal scare people
</blockquote></details>

I added the panic just because this code path should never be hit since we skip generating commands for any endpoint that takes a binary param. I could just remove it for now if you prefer.
___
*[`conjure/cliwriter.go` line 438 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0pusW01XfDqH5SiLLX:-N0qDCTvEMGfSvSwOtG1:b-4njd35) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/conjure/cliwriter.go#L438)):*

There's a `Unmarshal` method so you don't have to create a reader (unless this is future-proofing for reading from files...)
</blockquote></details>

Yeah, was planning to offer read from files for everything that uses this impl.
___
*[`conjure/snip/imports.go` line 62 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0poxRH7x74SJv39xLc:-N0qDU9r-O0BMY-8M86p:b-896fix) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/conjure/snip/imports.go#L62)):*

needs `.Clone`
</blockquote></details>

Done.
___
*[`conjure/snip/imports.go` line 78 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0pp4Sc7IBit42FX2le:-N0qDveP2LGpFr-U8Jfr:b-896fix) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/conjure/snip/imports.go#L78)):*

ioutil is deprecated, let's use `os.ReadFile`
</blockquote></details>

Done.
___
*[`integration_test/testgenerated/cli/cli-service.yml` line 16 at r3](https://reviewable.io/reviews/palantir/conjure-go/297#-N0ppxFo3gP5orO00UHH:-N0pyJf37GKgUA15pmgc:b-896fix) ([raw file](https://github.com/palantir/conjure-go/blob/ee592004837cdb4030fdca2eb769cea0c29741d6/integration_test/testgenerated/cli/cli-service.yml#L16)):*

nit: let's keep `rescue` out of this repo
</blockquote></details>

Done.

bmoylan (approved):
Approving so we can get this merged and start working with it, especially because it is opt-in. We're going to want full argument support (including binary and files) and more ergonomic client configuration, but this is a good start.

asanderson15:
Thanks, am planning to start implementing the remaining improvements this week time permitting.

Issue #298, opened 2022-04-22
Use generics to improve Ergonomics of generated union types


jamesross3:
### Problem
Today, Conjure-generated union types in Go are harder to handle than their counterparts, primarily due to the lack of generics. If I want to return a value with type `T` determined by visiting a union type, I currently have to either (1) define a stateful struct which implements the existing `Visitor` or `VisitorWithContext` interface, or (2) leverage closures by instantiating a variable outside of the AcceptFuncs method and setting it in the Visit methods provided to AcceptFuncs. These both have disadvantages; the former approach makes reuse of the stateful struct a recipe for concurrency bugs, and the latter forces the use of closures, which can degrade readability.

### Proposed Solution
The advent of generics in go1.18 lets us generate type-parametrized code to make usage much cleaner. Given the following union type definition:
```
MyUnion:
  union:
    stringVal: string
    boolVal: bool
```
I propose generating this code:
```
type MyUnionWithT[T any] MyUnion

func (u myUnionWithT) Accept(ctx context.Context, v myUnionVisitorWithT[T]) (T, error) {
  switch u.typ {
    case default:
      if u.typ == "" { return nil, fmt.Errorf("unknown")}
      return v.VisitUnknown(ctx, u.typ)
    case "stringVal":
      return v.VisitStringVal(ctx, *u.stringVal)
    case "boolVal":
      return v.VisitBoolVal(ctx, *u.boolVal)
  }
}

type MyUnionVisitorWithT[T any] interface {
	VisitStringVal(ctx context.Context, v string) (T, error)
	VisitBoolVal(ctx context.Context, v bool) (T, error)
	VisitUnknown(ctx context.Context, typ string) (T, error)
}
```

Where example usage looks like:
```
func Stringify(ctx, context.Context, myUnion MyUnion) (string, error) {
  var stringifyingVisitor MyUnionVisitorWithT[string] = NewStringifyingVisitor()
  return MyUnionWithT[string](myUnion).Accept(ctx, stringifyingVisitor)
}
```

jamesross3:
Draft: https://github.com/palantir/conjure-go/pull/300

tabboud:
Completed as of https://github.com/palantir/conjure-go/pull/300

PR #300, opened 2022-04-22
Propose generic visitor interfaces for union types

jamesross3:
draft proposal for https://github.com/palantir/conjure-go/issues/298

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/300)


conjure/unionwriter_test.go
```diff
@@ -0,0 +1,92 @@
...
+		{
+			Name: "stringVal", Type: types.String{},
+		},
+		{
+			Name: "boolVal", Type: types.Boolean{},
+		},
+	},
+}
+
+func TestUnionWriter_unionVisitorInterfaceT(t *testing.T) {
```
jamesross3, line 38: This is the file of interest in terms of seeing the code the new methods generate - once we agree on what we want these to look like, I will go ahead and move from draft to formal PR where the tested methods are actually invoked.

nmiyake: Would you mind adding some integration tests or a branch somewhere that has an example of using the generated code? I think that being able to see the actual generated code would be helpful in doing the review.

nmiyake (changes requested):
Per the note in the PR, can you include an example of the usage of the generated code, either as part of the PR or as a link to code that exists somewhere else?

Also, the generated code here would only be compatible with projects that use Go 1.18. This should be the majority of projects, but it isn't 100%. Do you have thoughts on the roll-out approach for this? Just proceeding and blocking upgrades unless a project moves to 1.18 is an approach that we could use, but whatever it is just want to make sure that we're explicit about it.

conjure/unionwriter_test.go
jamesross3: Added a basic example of usage here: https://github.com/jamesross3/conjure-go/blob/jr/union-generics-example-usage/conjure-api/conjure/spec/unions.conjure_test.go

The code that actually gets generated is now present in the latest commit ( in the file `conjure-api/conjure/spec/unions.conjure.go`)

jamesross3:
> Per the note in the PR, can you include an example of the usage of the generated code, either as part of the PR or as a link to code that exists somewhere else?
> 
> Also, the generated code here would only be compatible with projects that use Go 1.18. This should be the majority of projects, but it isn't 100%. Do you have thoughts on the roll-out approach for this? Just proceeding and blocking upgrades unless a project moves to 1.18 is an approach that we could use, but whatever it is just want to make sure that we're explicit about it.

(1) linked to a basic example (https://github.com/jamesross3/conjure-go/blob/jr/union-generics-example-usage/conjure-api/conjure/spec/unions.conjure_test.go) - happy to discuss or provide more examples if this isn't sufficient.
(2) Yes, I think simply blocking upgrades to this version of the plugin for any project on go 1.17 or earlier is reasonable at this point. I'm unsure how this may affect the mechanics of the excavator typically responsible for bumping the plugin.

conjure/conjure.go
```diff
@@ -69,10 +69,14 @@ func GenerateOutputFiles(conjureDefinition spec.ConjureDefinition, cfg OutputCon
...
 		if len(pkg.Unions) > 0 {
 			unionFile := newJenFile(pkg, def)
+			go1Dot18UnionFile := newJenFile(pkg, def)
+			go1Dot18UnionFile.Comment("//go:build go1.18")
 			for _, union := range pkg.Unions {
 				writeUnionType(unionFile.Group, union, cfg.GenerateFuncsVisitor)
+				writeUnionTypeWithGenerics(go1Dot18UnionFile.Group, union)
 			}
 			files = append(files, newGoFile(filepath.Join(pkg.OutputDir, "unions.conjure.go"), unionFile))
+			files = append(files, newGoFile(filepath.Join(pkg.OutputDir, "unions.conjure_go1.18.go"), go1Dot18UnionFile))
```
nmiyake: Can we change this to something like `unions_generics.conjure.go` or `unions_go118.conjure.go`?

For consistency, I'd like to maintain the `.conjure.go` suffix for all Conjure-generated files, and would also like to avoid using `go1.18` as part of the name since I'd like to reserve the dot (`.`) character for semantic delineation of path parts.

conjure/unionwriter_test.go
nmiyake: Thanks, seeing the example was very helpful!

nmiyake (changes requested):
Thanks for adding this functionality, this looks great!

Main point of feedback is around the name of the newly created/generated file, but otherwise the content looks good.

conjure/conjure.go
jamesross3: Makes sense, done - I chose `unions_generics.conjure.go`

nmiyake (approved):
This looks great, thanks!

PR #301, opened 2022-04-25
Update CI to publish using GitHub actions and use .palantir/go-version file

nmiyake:
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/301)


.circleci/config.yml
```diff
@@ -1,145 +1,287 @@
...
-  dist:
-    <<: *darwin-linux-no-cgo
-    steps:
-      - checkout
-      - *go-version
-      - *godel-cache-restore
-      - *godel-version
-      - *godel-cache-save
-      - run: ./godelw dist
+      - run:
```
nmiyake: This section is mostly lifted from the orb. This job needs to be custom because the executor has multiple docker images (includes `palantirtechnologies/conjure-verification-server:0.18.5` for tests). Otherwise, this could just the `godel/tests` job with the tag value of `conjure-verifier`.

I'm not sure that it's possible to manually define an executor that uses multiple Docker images -- if it is, then it's possible that we could leverage that instead. The other approach would be to define an orb action that basically fills in all the pre/post steps, and then we could just use that and drop in the custom logic for running tests and setting up the executor.

nmiyake: Actually maybe that is possible... let me try this out.

nmiyake: Huh, that actually worked -- hooray for rubber-ducking!

tabboud (approved):

PR #305, opened 2022-05-04
Improvement: Add verbose mode to generated CLIs

asanderson15:
==COMMIT_MSG==
Add verbose mode to generated CLIs
==COMMIT_MSG==

* For generated CLI commands, only initialize a logging context if the `-v` or `--verbose` flag is set. This keeps output clean so that it can be piped through to other commands or scripts.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/305)

### DIFF
--- a/changelog/@unreleased/pr-305.v2.yml
+++ b/changelog/@unreleased/pr-305.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add verbose mode to generated CLIs
+  links:
+  - https://github.com/palantir/conjure-go/pull/305
--- a/conjure/cliwriter.go
+++ b/conjure/cliwriter.go
@@ -32,6 +32,7 @@ const (
 
 	bearerTokenFlagName = "bearer_token"
 	confFlagName        = "conf"
+	verboseFlagName     = "verbose"
 )
 
 var (
@@ -40,6 +41,7 @@ var (
 	cliReservedArgNames = []string{
 		confFlagName,
 		bearerTokenFlagName,
+		verboseFlagName,
 	}
 )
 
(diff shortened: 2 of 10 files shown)


bmoylan (changes requested):
> Reviewed 8 of 8 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/305)* status: all files reviewed, 1 unresolved discussion (waiting on @asanderson15)

___
*[`conjure/cliwriter.go` line 122 at r2](https://reviewable.io/reviews/palantir/conjure-go/305#-N1Fwuyr9iLpKFFHpi-L:-N1Fwuyr9iLpKFFHpi-M:b6wc902) ([raw file](https://github.com/palantir/conjure-go/blob/43ffb42fb75619217293053cad6577e6f8873bf5/conjure/cliwriter.go#L122)):*
> ```Go
> 	file.List(jen.Id("verbose"), jen.Err()).Op(":=").Id("flags").Dot("GetBool").Call(jen.Lit(verboseFlagName))
> 	file.If(jen.Op("!").Id("verbose").Op("||").Err().Op("!=").Nil()).Block(
> 		jen.Return(jen.Id("ctx")))
> ```

instead should we initialize a noop/discard logger? Otherwise I think the default behavior will kick in and still print things

asanderson15 (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/305)* status: 1 of 10 files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[`conjure/cliwriter.go` line 122 at r2](https://reviewable.io/reviews/palantir/conjure-go/305#-N1Fwuyr9iLpKFFHpi-L:-N1KezZ98PetPSD4BQLy:b-7q1dgo) ([raw file](https://github.com/palantir/conjure-go/blob/43ffb42fb75619217293053cad6577e6f8873bf5/conjure/cliwriter.go#L122)):*

instead should we initialize a noop/discard logger? Otherwise I think the default behavior will kick in and still print things
</blockquote></details>

Yeah, good call. I forgot about the default impls. Updated.

bmoylan (approved):
> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/305)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @asanderson15)

PR #306, opened 2022-05-06
Improvement: Generated CLIs support input via file, base64, and STDIN

asanderson15:
==COMMIT_MSG==
Add generated CLI support for:
- binary args via file, base64-encoded strings, and stdin
- collection and complex types support input from file and stdin (in addition to existing json-encoded strings)
==COMMIT_MSG==

* We now generate commands for endpoints that take binary arguments
* Binary arguments may be input via file, STDIN, or b64 encoded strings
* We also add support for file and STDIN input for collection and other complex type arguments

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/306)


conjure/cliwriter.go
```diff
@@ -436,17 +432,60 @@ func astForEndpointCollectionParam(file *jen.Group, argName string, flagVar jen.
...
+			jen.Id(argReaderName),
 			jen.Op("&").Id(argName),
 		),
 		jen.Err().Op("!=").Nil(),
-	).Block(jen.Return(snip.WerrorWrapContext().
-		Call(jen.Id("ctx"), jen.Err(), jen.Lit(fmt.Sprintf("invalid value for %s argument", param.Name)))))
+	).Block(
+		jen.Return(snip.WerrorWrapContext().
+			Call(jen.Id("ctx"), jen.Err(), jen.Lit(fmt.Sprintf("invalid value for %s argument", param.Name)))))
+	file.Id(argReaderName).Dot("Close").Call()
```
tabboud: We may want to move this above the json decoding and turn it into a `defer bodyArgReader.Close()` in the case that the JSON decoding fails. This may be a non-issue if Go automatically closes file descriptors when the program ends, but I am not sure of the answer atm.

integration_test/testgenerated/cli/cli_test.go
```diff
@@ -121,6 +123,24 @@ func TestCommand_EchoCustomObject(t *testing.T) {
...
+		args := []string{
+			"",
+			"echoCustomObject",
+			"--body",
+			"@./testdata/echoCustomObject",
+		}
+		t.Run("success", func(t *testing.T) {
+			client, testServiceCommand := getMockClientAndTestCommand()
+			client.On("EchoCustomObject", mock.Anything, &customObject).Return(&customObject, nil).Times(1)
+			executeAndAssertSuccessAndOutput(t, testServiceCommand, args, client, "{\n    \"data\": \"Ynl0ZXM=\"\n}\n")
```
tabboud, line 143: I'm not the biggest fan of the testdata file + hardcoded expectations as it will eventually fall out of date if new features are added. I think a better approach would be to inline the expected file content and write them to a temp file for tests that need to read from a file. The only caveat being that the final assertion of expected output becomes a bit trickier since you have to work around the formatting differences.

With all that said, I am happy to move forward with this approach and we can always inline things down the road. I don't expect there to be a ton of features added here so it's probably safe enough.

changelog/@unreleased/pr-306.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: improvement
+improvement:
+  description: |-
+    Add generated CLI support for:
```
tabboud: Rather than state what was done, can you re-word this to convey what has changed that an end user would care about if they read the release notes.

For example, instead of just saying "Add generated CLI support for ...", you can say that binary args can now be provided to generated CLIs via a file (using the @ notation), stdin (using the "@-" notation), etc.

tabboud (commented):
Overall logic looks good to me, just a few comments.

integration_test/testgenerated/cli/cli_test.go
asanderson15: Done.

changelog/@unreleased/pr-306.v2.yml
asanderson15: Done.

conjure/cliwriter.go
asanderson15: yep, that's a good call  - done

tabboud (approved):

PR #406, opened 2023-01-23
Ensure that imports are not removed in generated code

nmiyake:
Makes it such that the "ptimports.Process" call that is called for individual generated code files does not add or remove imports and then adds a step where "ptimports.Process" is called on all generated code files after they are written. Fixes an issue where generated code would not include imports that referenced packages where the last component of the package path was of the form "v[0-9]+".

Fixes #405

## Before this PR
See #405 for details, but before this PR, generating code for Conjure definitions that have packages where the last component is of the form `v[0-9]+` (for example, "com.palantir.foo.v2") would not generate the import statement needed to compile (so the first run of the generator would produce code that didn't compile). Running the generator a second time would add the expected imports.

## After this PR

==COMMIT_MSG==
Fixes an issue where generated code would not include imports that referenced packages where the last component of the package path was of the form "v[0-9]+".
==COMMIT_MSG==

## Possible downsides?
This fix adds a step where, after all the generated files are written, they are all formatted and written again. This process adds computational overhead on the order of the number of files written, but I don't think it should be a major concern as the operation is not too expensive and generating the files is not a hot path (it's an operation that is expected to take a decent amount of time and is run in a one-off manner in workflows).

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/406)

Issue #405, opened 2023-01-23
Generated code does not compile if Conjure definition references type in a package where last element is of the form "v[0-9]+"


nmiyake:
## What happened?
When I generated Conjure output files for a definition that defined a type in a package that ended in "v2" (`com.palantir.test.v2`) and an object in a different package that referenced that type, the generated code did not initially compile.

Running the generator again (after the initial non-compiling code was generated) added the proper import statement, after which the code compiled as expected.

The initial behavior is incorrect (running the generator for a valid definition should create valid code), and the fact that running the same Conjure generation operation again generates different code that fixes the issue is unintuitive (the generator is expected to be idempotent).

----------

Example Conjure definition:

```
types:
  definitions:
    objects:
      ObjectOne:
        package: com.palantir.test.common
        fields:
          objectTwo:
            type: ObjectTwo
      ObjectTwo:
        package: com.palantir.test.v2
        fields:
          name:
            type: string
```

IR for the definition:

```
{
  "version" : 1,
  "errors" : [ ],
  "types" : [ {
    "type" : "object",
    "object" : {
      "typeName" : {
        "name" : "ObjectOne",
        "package" : "com.palantir.test.common"
      },
      "fields" : [ {
        "fieldName" : "objectTwo",
        "type" : {
          "type" : "reference",
          "reference" : {
            "name" : "ObjectTwo",
            "package" : "com.palantir.test.v2"
          }
        }
      } ]
    }
  }, {
    "type" : "object",
    "object" : {
      "typeName" : {
        "name" : "ObjectTwo",
        "package" : "com.palantir.test.v2"
      },
      "fields" : [ {
        "fieldName" : "name",
        "type" : {
          "type" : "primitive",
          "primitive" : "STRING"
        }
      } ]
    }
  } ],
  "services" : [ ],
  "extensions" : {}
}
```

Initial lines of output of first run (at `common/structs.conjure.go`):

```
// This file was generated by Conjure and should not be manually edited.

package common

import (
        "github.com/palantir/pkg/safejson"
        "github.com/palantir/pkg/safeyaml"
)

type ObjectOne struct {
        ObjectTwo v2.ObjectTwo `json:"objectTwo"`
}
```

Initial lines of output of second run (at `common/structs.conjure.go`):

```
// This file was generated by Conjure and should not be manually edited.

package common

import (
        v2 "github.com/palantir/conjure-go/v6/cmd/outdir/test/v2"
        "github.com/palantir/pkg/safejson"
[...]

PR #408, opened 2023-01-30
Revert "Ensure that imports are not removed in generated code (#406)"

nmiyake:
This reverts commit ffd6daaaee62dd944e359df43db7f2abab7333ec.

## Before this PR
PR #406 changed the behavior of the generator such that required imports would always be properly added by `conjure.Generate`. However, this made it such that the output of `conjure.Generate` and `conjure.GenerateOutputFiles` does not always match -- in cases where they don't, this causes the "verify" operation of godel-conjure-plugin to fail.

## After this PR

==COMMIT_MSG==
Reverts the conjure-go behavior change introduced in v6.36.0 to match previous versions: certain named imports generated by Conjure code will only be generated if the Conjure generator is run twice, but once this is done, the output of `conjure.Generate` and `conjure.GenerateOutputFiles` will match, which prevents verification issues.
==COMMIT_MSG==

## Possible downsides?

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/408)


nmiyake:
@bmoylan sorry, after this merged and I tested it on a target repo, found that it breaks the "verify" operation. I've documented the issue at https://github.com/palantir/conjure-go/issues/405#issuecomment-1409139133 so that we know to consider this case for the next attempt at the fix.

In the meantime, would like to revert the behavior, since the manner in which the previous misbehavior manifested is easier to work around.

bmoylan (approved):

PR #409, opened 2023-02-01
fix: Force import alias when package ends in /v[0-9]+$

bmoylan:
This is a more targetted fix for #405 than #406 reverted in #408.

Rather than operate on the generated files, we predict invalid package names and force an import alias. I've added a second test type to demonstrate how jennifer [deconflicts](https://github.com/dave/jennifer/blob/72bc3ba92d8262c9930a78584fcf0e4d3b54009c/jen/file.go#L199) aliases.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/409)


conjure/conjure.go
```diff
@@ -129,3 +134,7 @@ func newGoFile(filePath string, file *jen.File) *OutputFile {
 		file:    file,
 	}
 }
+
+func packageSuffixRequiresAlias(importPath string) bool {
+	return regexp.MustCompile(`/v[0-9]+$`).MatchString(importPath)
```
nmiyake, line 139: This looks great, thanks! The implementation the Go team uses for this is at https://github.com/palantir/go-ptimports/blob/93cb0022ccc2ca0a21b3ac3396b85349567cc40a/vendor/golang.org/x/tools/internal/imports/fix.go#L1145-L1153, but I think this regexp matching is more readable and fine.

nmiyake (approved):
Thanks, this looks great!

The other scenario that seems to be special-cased in the Go code is packages that start with `go-`, but I don't think that's a concern/case that we need to handle for Conjure since I'm not sure that packages are allowed to have a hyphen in the name (and even if they do, we probably have some safety from a practical perspective in that such package names should likely be rare).

PR #432, opened 2023-03-02
render conjure deprecation warnings in generated code

bwaldrep:
## Before this PR
Conjure allows marking endpoints, fields, and enum values as deprecated, with a message indicating what users should migrate to. Conjure-go does not render these deprecation notices in generated client code, making it more difficult to notice when you are using deprecated endpoints or to easily understand what you should use instead.

## After this PR

==COMMIT_MSG==
 - Conjure-go will unmarshal EnumValues marked as deprecated without throwing, in accordance with the [spec](https://github.com/palantir/conjure/blob/master/docs/spec/conjure_definitions.md#enumvaluedefinition)
 - Deprecation warnings will be rendered if defined on endpoints, enum values, and fields
 - The repo will build on arm macbooks
==COMMIT_MSG==

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/432)

This should be a docs-only change to existing generated code, making this low risk.

### DIFF
--- a/changelog/@unreleased/pr-432.v2.yml
+++ b/changelog/@unreleased/pr-432.v2.yml
@@ -1,0 +1,8 @@
+type: improvement
+improvement:
+  description: |-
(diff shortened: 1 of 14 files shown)


bmoylan (approved):
thanks!

PR #441, opened 2023-03-13
improvement: Add graph types and utils

splucs:
## Summary
Adding graph types and utils that will be used to implement https://github.com/palantir/conjure-go/tree/develop/conjure/cycles

## Changelog
==COMMIT_MSG==
Add graph types and utils
==COMMIT_MSG==

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/441)


godel/config/check-plugin.yml
```diff
@@ -4,3 +4,4 @@ checks:
     - value: should have comment or be unexported
     - value: or a comment on this block
     - value: struct field Uid should be UID
+    - value: should be consistent with previous receiver name
```
splucs, line 7: This check is failing because the plugin is not working with generics and thinks that `graph[T]` is the same as `node[T]`.

jamesross3 (approved):

PR #445, opened 2023-03-15
improvement: Add bitset data structure for fast set<int> merges

splucs:
## Summary
This PR adds the bitset data structure that implements `set<int>` with `O(1)` operations and `O(n^2)` with a multiplicative factor of `1/64` merges at the exchange of always using `O(n)` size, where `n` is the maximum size of the bitset (bitset can only contain numbers in `[0,n-1]`). This is fine because we expect this to be the case for the partition algorithm described in https://github.com/palantir/conjure-go/tree/develop/conjure/cycles

## Changelog
==COMMIT_MSG==
Add bitset data structure for fast set<int> merges
==COMMIT_MSG==

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/445)


jamesross3 (approved):
> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/conjure-go/445)* status: 2 of 4 files reviewed, all discussions resolved

PR #471, opened 2023-05-27
fix: Correctly break import cycles if type graph is disconnected

splucs:
## Summary
The previous algorithm ran under the assumption that all package cycles happened within a connected component of the type graph, which is not true. When SCCs that belong to distinct connected components of the type graph are merged, two disconnected DAGs become connected and the assumption that we can just merge on depth no longer works.

This PR changes the partitioning algorithm significantly to attempt a greedy best effort of merging. We can no longer guarantee that the number of new Go packages is minimal :(.

In addition, because we now have to run a new dfs for each merging SCC to update the dependencies and disallowed graph, the complexity increases from `O(n^2+m)` to `O(n^3+m)`, but the new dfs is parse enough that I don't believe this will be a problem.

## Testing
Added a new test case that reproduces the issue that the previous algorithm couldn't catch.

## Changelog
==COMMIT_MSG==
Correctly break import cycles if type graph is disconnected
==COMMIT_MSG==

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/471)


conjure/cycles/README.md
```diff
@@ -218,15 +215,39 @@ the merging sets being the bottleneck. We can, however, use the [bitset](https:/
...
+    If it can, put `su` in it and stop iterating over buckets.
+  - If there are none, create a new bucket for `su`'s package set and put `su` in it.
+- When two SCCs are merged, we must update the `dependencies` and `disallowed` sets of all SCCs that can
+  reach the merged nodes. This is done by traversing the reverse graph of the SCC graph and updating the sets using
+  the same recursive algorithm as above.
+
+The algorithm above runs in worst case `O(n^3 + m)` as a dfs on the reverse SCC graph is required for each SCC to
+recompute the `dependencies` and `disallowed` sets of all SCCs that can reach the merging SCC.
+The bitset optimization also applies here, which enables the algorithm to run with a multiplicative factor of `1/64`.
+An important node is that when nodes are merged, new edges are introduced in the SCC graph, which means that `m` can
```
nmiyake: Typo? "An important node" -> "An important note"?

conjure/cycles/graph_test.go
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2018 Palantir Technologies. All rights reserved.
```
nmiyake: Since this is a new file, year should be 2023 (the `./godelw verify` should automatically add the current year if the package has no comment/header, but I'm assuming this was copy-pasted)

nmiyake (commented):
Thanks! Very minor feedback on possible comment typo and copyright header, but content looks good.

nmiyake (approved):
Thanks!

PR #524, opened 2023-10-17
improvement: Union types check that variant is non-nil to avoid panic

### LINKED ISSUE #523: Generated union code panics when marshalling union type with `nil` member
## What happened?
Generate Conjure using the following definition:

```
types:
  definitions:
    default-package: com.palantir.apollo.catalog.api.module
    objects:
      VariableSpec:
        union:
          string: StringVariable

      StringVariable:
        fields:
          defaultValue:
            type: optional<string>
            safety: unsafe
          description:
            type: string
            safety: unsafe
```

(from https://github.pb/foundry/apollo-catalog/blob/develop/apollo-catalog-api/src/main/conjure/ModuleService.yml#L107-L118)

This generates the following Go:

```go
type VariableSpec struct {
	typ    string
	string *StringVariable
}

type variableSpecDeserializer struct {
	Type   string          `json:"type"`
	String *StringVariable `json:"string"`
}

func (u *variableSpecDeserializer) toStruct() VariableSpec {
	return VariableSpec{typ: u.Type, string: u.String}
}

func (u *VariableSpec) toSerializer() (interface{}, error) {
	switch u.typ {
	default:
		return nil, fmt.Errorf("unknown type %s", u.typ)
	case "string":
		return struct {
			Type   string         `json:"type"`
			String StringVariable `json:"string"`
		}{Type: "string", String: *u.string}, nil
	}
}
```

Note the `String: *u.string` at line 267 (https://github.pb/deployability/apollo-cli/blob/c66d938a2da8d1ec054f9d8a84ab21c0f08f5795/internal/generated/conjure/apollo_catalog_api/apollo/catalog/api/module/unions.conjure.go#L267) -- because this is a dereference of `u.string`, if `StringVariable` is `nil`, this will panic.

Although it's not possible to construct a `nil` value via constructors, it is via JSON unmarshal. Thus, the following code panics:

```
	var spec module.VariableSpec

	err := json.Unmarshal([]byte(`{"type": "string"}`), &spec)
	require.NoError(t, err)

	_, err = spec.MarshalJSON()
	require.NoError(t, err)
```

Either:

1. If `StringValue` cannot be `nil` in the Union type, then the `json.Unmarshal` call should return an error
2. Otherwise, the `MarshalJSON` code should produce valid JSON or produce an error (but not panic)

From my reading of the code, it seems like (1) is more correct -- in the union type definition, the `StringVariable` definition is not optional, so it being `nil` should not be allowed (the expected unmarshal should produce a `VariableSpec` with a non-`nil` `StringVariable` that has `nil`/empty content

bmoylan:
Fixes #523

This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/524)

### DIFF
--- a/conjure/unionwriter.go
+++ b/conjure/unionwriter.go
@@ -108,6 +108,20 @@ func writeUnionType(file *jen.Group, unionDef *types.UnionType, genAcceptFuncs b
 			jen.Err().Op("!=").Nil(),
 		).Block(jen.Return(jen.Err())),
 		jen.Op("*").Id(unionReceiverName).Op("=").Id("deser").Dot("toStruct").Call(),
+		jen.Switch(jen.Id(unionReceiverName).Dot("typ")).BlockFunc(func(cases *jen.Group) {
+			cases.Default().Block(jen.Return(
+				snip.FmtErrorf().Call(jen.Lit("unknown type %s"), jen.Id(unionReceiverName).Dot("typ"))))
+			for _, fieldDef := range unionDef.Fields {
+				cases.Case(jen.Lit(fieldDef.Name)).BlockFunc(func(caseBody *jen.Group) {
+					if !fieldDef.Type.IsOptional() {
+						selector := jen.Id(unionReceiverName).Dot(transforms.PrivateFieldName(fieldDef.Name))
+						caseBody.If(selector.Op("==").Nil()).Block(
+							jen.Return(snip.FmtErrorf().Call(jen.Lit(fmt.Sprintf("field %s is required", fieldDef.Name)))),
+						)
+					}
+				})
+			}
+		}),
 		jen.Return(jen.Nil()),
 	))
 
--- a/conjure/unionwriter.go
+++ b/conjure/unionwriter.go
@@ -15,6 +15,7 @@
 package conjure
 
 import (
+	"fmt"
 	"github.com/dave/jennifer/jen"
 	"github.com/palantir/conjure-go/v6/conjure/snip"
 	"github.com/palantir/conjure-go/v6/conjure/transforms"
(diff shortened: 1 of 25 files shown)


nmiyake (commented):
Thanks for the PR!

Looking at the implementation/generated code, it looks like the approach here is to update the visitor `AcceptFuncs` and the JSON marshal `toSerializer` to perform a `nil` check so that we error instead of panic, correct?

This is definitely strictly better than the current state, but the most correct fix would be to flag/fail during `Unmarshal`, right? As it currently stands, if we unmarshal an invalid union type of this form, it may stick around for a while and only throw an error on visit/marshal, which could be far removed from the unmarshal call that created the object.

Even if we don't want to mess with the actual internals of the unmarshal itself, from my reading of the generated code, it looks like we could add a type switch on `u.typ` (like the one in `toSerializer`) to the `toStruct` method with the same "field required" checks, in which case we would properly throw the error on unmarshal (we could still keep the current checks just as added safety, but adding the check on unmarshal seems more robust)

conjure/unionwriter.go
```diff
@@ -106,6 +108,18 @@ func writeUnionType(file *jen.Group, unionDef *types.UnionType, genAcceptFuncs b
 			jen.Err().Op("!=").Nil(),
 		).Block(jen.Return(jen.Err())),
 		jen.Op("*").Id(unionReceiverName).Op("=").Id("deser").Dot("toStruct").Call(),
+		jen.Switch(jen.Id(unionReceiverName).Dot("typ")).BlockFunc(func(cases *jen.Group) {
+			for _, fieldDef := range unionDef.Fields {
+				cases.Case(jen.Lit(fieldDef.Name)).BlockFunc(func(caseBody *jen.Group) {
+					if !fieldDef.Type.IsOptional() {
+						selector := jen.Id(unionReceiverName).Dot(transforms.PrivateFieldName(fieldDef.Name))
+						caseBody.If(selector.Op("==").Nil()).Block(
+							jen.Return(snip.FmtErrorf().Call(jen.Lit(fmt.Sprintf("field %s is required", fieldDef.Name)))),
```
nmiyake: Looking at generated code, I wonder if error would be more readable if field name was quoted instead (basically have a `fmt.Sprintf` call with `%q` as formatter and `fieldDef.Name` as value)

conjure/unionwriter.go
```diff
@@ -74,7 +76,7 @@ func writeUnionType(file *jen.Group, unionDef *types.UnionType, genAcceptFuncs b
 				jen.Nil(), snip.FmtErrorf().Call(jen.Lit("unknown type %s"), jen.Id(unionReceiverName).Dot("typ"))))
```
nmiyake: If we quote the field name, then would likely want to do it for the "unknown type" output here as well

nmiyake (approved):
Thanks! Changes and generated code look good, and verified locally that this fixes the specific case I encountered as well.

Question/suggestion about quoting the invalid field name in the error for readability, but not a major issue, so approving PR.

PR #565, opened 2024-02-23
Fix issue with cycle detection and 

DHerls:
## Before this PR
The cycle detection algorithm did not check for disallowed nodes in the same merged group as the target node. 

### Example:
Nodes: 0, 1, 2, and 3

Colors:
* 0 - Red
* 1 - Red
* 2 - Blue
* 3 - Red

Edges:
* 2 -> 1
* 3 -> 2

### Initial disallowed state
```
0: []
1: []
2: [1]
3: [2, 1]
```
### Partitioning
1. 0 and 1 attempt to merge
1. 0 and 1 are the same color and are not on each others' disallow lists so they merge
1. 3 and 0 attempt to merge
1. 3 and 0 are the same color and are not on each others' disallow lists so they merge
1. 3 and 1 are now in the same group, causing a cycle

## After this PR

1. The algorithm now keeps track of all of the nodes that have been merged with a given node
1. When evaluating whether two nodes can merge, check whether there is any intersection between each node's disallowed nodes and the other node's merged nodes

==COMMIT_MSG==
Handle case in cycle detection where a node's previously merged nodes were not checked when evaluating whether it could merge with another node
==COMMIT_MSG==

## Possible downsides?
* Intersection is more expensive
* We are adding another bitset for every node
* This may cause previously generated conjure bindings to change if there were undetected cycles

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/565)

### DIFF
--- a/changelog/@unreleased/pr-565.v2.yml
+++ b/changelog/@unreleased/pr-565.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Handle case in cycle detection where a node's previously merged nodes
+    were not checked when evaluating whether it could merge with another node
+  links:
+  - https://github.com/palantir/conjure-go/pull/565
--- a/conjure/cycles/bitset.go
+++ b/conjure/cycles/bitset.go
@@ -72,3 +72,12 @@ func (bs bitset) merge(o bitset) bitset {
 	}
 	return ret
 }
+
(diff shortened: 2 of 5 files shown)


splucs (approved):
Thanks for fixing this!

PR #620, opened 2024-09-11
improvement: Upgrade CGR and use httpclient.RequestBody for binary uploads

bmoylan:
Uses https://github.com/palantir/conjure-go-runtime/pull/695

This PR changes the type of `binary` body arguments from `func() io.ReadCloser` to `httpclient.RequestBody`, allowing callers to more accurately specify how the body should be streamed and whether it can be retried.

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/620)


go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/dave/jennifer v1.4.2-0.20211112003305-45cc0b7eb71a
 	github.com/julienschmidt/httprouter v1.3.0
 	github.com/nmiyake/pkg/dirs v1.1.0
-	github.com/palantir/conjure-go-runtime/v2 v2.83.0
+	github.com/palantir/conjure-go-runtime/v2 v2.83.1-0.20241107181000-fb10c9ca7ad8 // Requires https://github.com/palantir/conjure-go-runtime/pull/695
```
nmiyake: We should cut a release for this before proceeding with this PR

nmiyake (changes requested):
Thanks! Change looks good, but would like to wait for a release so we can reference a release version of a module in `go.mod` before proceeding

go.mod
bmoylan: @nmiyake there's now a CGR release so this should be ready for review!

nmiyake (approved):

PR #626, opened 2024-10-31
Fix: Update local type alias naming scheme to prevent conflicts with object field types

Arthur-Befumo:
## Before this PR
Prior to this PR, the following valid conjure object definition would result in invalid generated code:
```
// Conjure def
UserName:
    alias: string
UserAlias:
    alias: string
User:
    fields:
        name: UserName
        aliases: list<UserAlias>

// Generated UnmarshalJSON code
func (o *User) UnmarshalJSON(data []byte) error {
    type UserAlias User
    var rawUser UserAlias
    if err := safejson.Unmarshal(data, &rawUser); err != nil {
        return err
    }
    if rawUser.UserAliases == nil {
        // This line is invalid, since the original UserAlias type has been 'replaced' by an alias for the User type
        rawUser.UserAliases = make([]UserAlias, 0)
    }
    *o = User(rawDomain)
    return nil
}
```

## After this PR
The local alias for an object with fields that are collections is now named `"_tmp" + object.Name`, ensuring that it will not conflict with any of the collection types since those will always be exported. A lone underscore would be sufficient, but I feel like the `tmp` makes the intent clearer.

==COMMIT_MSG==
Update local type alias naming scheme to prevent conflicts with object field types
==COMMIT_MSG==

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/626)


bmoylan:
@Arthur-Befumo can you update from develop to fix the failing test?

bmoylan (dismissed):
lgtm but you need to run `./godelw license`

bmoylan (approved):

PR #629, opened 2024-11-01
improvement: Build with go 1.22.8 and godel 2.121.0

bmoylan:
- Upgrade go to 1.23.2
- Incorporate #624 to support newer runtime
- Update witchcraft-go-server to 2.87 to fix tests (picking up https://github.com/palantir/witchcraft-go-server/pull/837)
- Add `circle-all` job to placate new protected branch configuration

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/629)


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/palantir/conjure-go/v6
 
-go 1.21
+go 1.23
+
+toolchain go1.23.2
```
nmiyake: Can you remove this line? We generally don't want to enforce a `toolchain`

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/palantir/conjure-go/v6
 
-go 1.21
+go 1.23
```
nmiyake: Can you update this to `go 1.23.0`? Newest format allows for full specification of version

nmiyake (changes requested):

go.mod
bmoylan: done!

go.mod
bmoylan: done!

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/palantir/conjure-go/v6
 
-go 1.21
+go 1.23.2
```
nmiyake: If we want this to be 1.23, then this line should be `1.23.0`.

The idea is that we want it to be the first release of the major version so that the module can use the language features, but also want to be as permissive as possible for vendoring/use by other modules (we don't want to block import for a different module that declares `go 1.23.1`).

This line sets a *minimum* for the toolchain required to build it, but if the local go version is greater than this version, the module will still be built with the "newer" go toolchain (unless there is a different directive that forces a specific toolchain version)

nmiyake (changes requested):

nmiyake (changes requested):
With this setup, need to update `.palantir/go-version` to `go.1.22.8` as well

nmiyake (approved):

Issue #415, opened 2023-02-13
Runtime panic occurs if there are multiple Conjure errors with the same name


nmiyake:
## What did you do?
1. Publish the following Conjure definition (which will be referred to as `common-service.conjure.json`):

```
types:
  definitions:
    default-package: com.palantir.test
    errors:
      EntityNotFound:
        namespace: V6
        code: NOT_FOUND
        docs: Thrown when the provided entity ID is not found
        safe-args:
          entityId: string
```

2. Create module `github.com/palantir/test-library-module` that generates code using the `common-service.conjure.json` definition and references the generated code
3. Create module `github.com/palantir/test-service-module` that generates code using the `common-service.conjure.json` definition and references the generated code AND that has `github.com/palantir/test-library-module` as a module dependency in `go.mod`
4. Run the `github.com/palantir/test-service-module`

## What happened?
The program crashes on startup with the following error:

```
panic: ErrorName V6:EntityNotFound already registered as v6.EntityNotFound

goroutine 1 [running]:
github.com/palantir/conjure-go-runtime/v2/conjure-go-contract/errors.RegisterErrorType({0x10237d60e, 0x11}, {0x10272f050?, 0x1021e2960})
```

## What did you want to happen?
Program should start normally

nmiyake:
The issue is that the generated `errors.conjure.go` file generates the following code:

```
func init() {
	errors.RegisterErrorType("V6:EntityNotFound", reflect.TypeOf(EntityNotFound{}))
}
```

The `errors.RegisterErrorType` function uses shared storage and panics if registration is done with the same key.

In the steps above, the repository ends with with 2 copies of the same `errors.conjure.go` (one in the consumed/vendored module and one in the local module), so the `init` function is called twice, which causes the panic at runtime.

It is an intentional design decision to require that error types be globally unique, but this poses a problem in the above scenario. In such a scenario, currently the only workaround is to modify the generated code (basically, to remove the error registration). Although this works, it involves manual edits and also makes it such that typed errors aren't returned for the locally generated code.

## Possible Solutions

### Associate error registration with particular client
The purpose of registering the errors is to allow `conjure-go-runtime` to provide a strongly typed error -- it maps the unmarshal of the error to the type provided.

[...]

PR #636, opened 2024-11-26
Internal error registry per package

bmoylan:
Replace global conjure-type error registration with a generated registry per conjure definition.

Creates a new `internal/conjureerrors` package in each output directory for conjure definitions that include errors. Generated client implementations provide the definition-specific registry as a ConjureErrorDecoder to be used when deserializing a non-2xx JSON response.

Clients will no longer be able to deserialize errors not defined in their own conjure definition.

Generating multiple IRs into the same output directory will result in all definitions sharing the same error registry.

Requires https://github.com/palantir/conjure-go-runtime/pull/724

Addresses #415

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/636)


go.mod
```diff
@@ -3,10 +3,10 @@ module github.com/palantir/conjure-go/v6
 go 1.22.0
 
 require (
-	github.com/dave/jennifer v1.4.2-0.20211112003305-45cc0b7eb71a
+	github.com/dave/jennifer v1.7.1
 	github.com/julienschmidt/httprouter v1.3.0
 	github.com/nmiyake/pkg/dirs v1.1.0
-	github.com/palantir/conjure-go-runtime/v2 v2.86.0
+	github.com/palantir/conjure-go-runtime/v2 v2.86.1-0.20241126014203-f52ba8f67d54
```
bmoylan: Do not merge; requires https://github.com/palantir/conjure-go-runtime/pull/724

tabboud (commented):
Couple notes but otherwise this looks good to me:
- Not being able to deserialize errors not defined by a Clients conjure definition seems more than reasonable, and in fact maybe more correct.
- It is common for projects to generate all conjure into a shared output directory, but this would be a good reason to push projects into namespacing their conjure dependencies, which I have found to be a better UX anyways. Besides this behavior would be the same as what currently exists and would not require a migration for users to pick this up.

tabboud (approved):

PR #687, opened 2025-03-28
fix: Path params are ordered based on the path template, not endpoint definition

bmoylan:
Previously, `astForEndpointMethodBodyRequestParams` would create the `httpclient.WithPathf` parameter using the argument ordering present in the conjure definition/intermediate representation. This is incorrect. Instead, the path parameters must match their ordering in the endpoint's http path template.

We are NOT changing the argument ordering in the generated go interfaces, since it would be a (potentially silent) break and arguments are referenced by name, not position.

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/687)


cycles/testdata/cycle-within-pkg/conjure/com/palantir/services/services.conjure.go
```diff
@@ -35,7 +34,7 @@ func (c *myServiceClient) Endpoint1(ctx context.Context, authHeader bearertoken.
 	requestParams = append(requestParams, httpclient.WithRPCMethodName("Endpoint1"))
 	requestParams = append(requestParams, httpclient.WithRequestMethod("GET"))
 	requestParams = append(requestParams, httpclient.WithHeader("Authorization", fmt.Sprint("Bearer ", authHeader)))
-	requestParams = append(requestParams, httpclient.WithPathf("/endpoint1", url.PathEscape(fmt.Sprint(arg1Arg))))
+	requestParams = append(requestParams, httpclient.WithPathf("/endpoint1"))
```
bmoylan: this is an unintentional fix. Previously this was an invalid printf statement and now it has the correct number of pattern values (none).

conjure/servicewriter.go
```diff
@@ -345,9 +345,21 @@ func astForEndpointMethodBodyRequestParams(methodBody *jen.Group, endpointDef *t
...
-			args.Add(snip.URLPathEscape()).Call(snip.FmtSprint().Call(jen.Id(transforms.ArgName(param.ParamID))))
+		// arguments
+		pathParams := endpointDef.PathParams()
+		pathParamsByID := make(map[string]*types.EndpointArgumentDefinition, len(pathParams))
+		for _, param := range pathParams {
+			pathParamsByID[param.ParamID] = param
+		}
+		for _, match := range pathParamRegexp.FindAllString(endpointDef.HTTPPath, -1) {
+			paramID := match[1 : len(match)-1]
+			if param, ok := pathParamsByID[paramID]; ok {
```
nmiyake, line 359: This may be overkill/unnecessary, but for maximal safety, would it make sense to have a counter for "numMatched", increment it in this if statement, then panic after the loop if `numMatched != len(pathParams)`? This would protect against case where not all expected path params were found in the endpoint definition HTTP path (although it's possible/likely that structure of Conjure will make it such that this invariant will never occur)

nmiyake (approved):

PR #714, opened 2025-07-02
Generate log safety struct tags for safelogging lint

sonnysasaka:
Generate log safety struct tags for logger safety implementation

This implements the Go struct safety annotation support by adding
`safelogging` struct tags and comments to the generated Go structs.

```go
// safelogging:@DoNotLog
type SafetyTestObject struct {
    SafeField     SafeString   `safelogging:"@Safe" json:"safeField"`
    DoNotLogField SecretString `safelogging:"@DoNotLog" json:"doNotLogField"`
    NormalField   string       `json:"normalField"`
}
```

This will enable safelogging library to inspect struct field safety annotations
preventing security incidents.

## After this PR

==COMMIT_MSG==
Generate log safety struct tags for Go client safety implementation
==COMMIT_MSG==

## Possible downsides?

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/714)

### DIFF
--- a/conjure/errorwriter.go
+++ b/conjure/errorwriter.go
@@ -59,7 +59,8 @@ func writeErrorType(file *jen.Group, def *types.ErrorDefinition) {
 func astErrorInternalStructType(file *jen.Group, def *types.ErrorDefinition) {
 	allArgs := append(append([]*types.Field{}, def.SafeArgs...), def.UnsafeArgs...)
 	// Use object generator to create a struct implementing JSON encoding for the error.
-	writeObjectType(file, &types.ObjectType{Name: transforms.Private(def.Name), Fields: allArgs})
+	safetyCache := make(map[types.Type]spec.LogSafety)
+	writeObjectType(file, &types.ObjectType{Name: transforms.Private(def.Name), Fields: allArgs}, safetyCache)
 }
 
 // Declare New and Wrap constructors
--- a/conjure/objectwriter.go
+++ b/conjure/objectwriter.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 
 	"github.com/dave/jennifer/jen"
+	"github.com/palantir/conjure-go/v6/conjure-api/conjure/spec"
 	"github.com/palantir/conjure-go/v6/conjure/snip"
 	"github.com/palantir/conjure-go/v6/conjure/transforms"
 	"github.com/palantir/conjure-go/v6/conjure/types"
@@ -28,10 +29,32 @@ const (
 	dataVarName     = "data"
 )
 
-func writeObjectType(file *jen.Group, objectDef *types.ObjectType) {
+// logSafetyToAnnotation converts a LogSafety value to its corresponding safelogging annotation string
+func logSafetyToAnnotation(safety spec.LogSafety_Value) string {
+	switch safety {
+	case spec.LogSafety_SAFE:
+		return "@Safe"
+	case spec.LogSafety_UNSAFE:
+		return "@Unsafe"
+	case spec.LogSafety_DO_NOT_LOG:
+		return "@DoNotLog"
+	default:
+		panic("unhandled LogSafety value: " + safety)
+	}
+}
+
+func writeObjectType(file *jen.Group, objectDef *types.ObjectType, safetyCache map[types.Type]spec.LogSafety) {
 	// Declare struct type with fields
 	containsCollection := false // If contains collection, we need JSON methods to initialize empty values.
-	file.Add(objectDef.Docs.CommentLine()).Type().Id(objectDef.Name).StructFunc(func(structDecl *jen.Group) {
+
+	// Compute the overall safety of the struct and add comment-based annotation
+	overallSafety := computeObjectSafety(objectDef, safetyCache)
+	var safetyComment jen.Code = jen.Null()
+	if !overallSafety.IsUnknown() {
+		safetyComment = jen.Comment("safelogging:" + logSafetyToAnnotation(overallSafety.Value()))
+	}
+
+	structFunc := func(structDecl *jen.Group) {
 		for _, fieldDef := range objectDef.Fields {
 			jsonTag := fieldDef.Name
 			if _, isOptional := fieldDef.Type.(*types.Optional); isOptional {
@@ -45,12 +68,25 @@ func writeObjectType(file *jen.Group, objectDef *types.ObjectType) {
 				// double quotes instead.
 				fieldTags["conjure-docs"] = strings.Replace(strings.TrimSpace(string(fieldDef.Docs)), "`", `"`, -1)
 			}
+
+			// Add safety struct tag based on field's safety annotation or type safety (with recursive computation)
+			fieldSafety := getSafetyFromField(fieldDef, safetyCache)
+			if !fieldSafety.IsUnknown() {
+				fieldTags["safelogging"] = logSafetyToAnnotation(fieldSafety.Value())
+			}
 			if fieldDef.Type.Make() != nil {
 				containsCollection = true
 			}
 			structDecl.Add(fieldDef.Docs.CommentLineWithDeprecation(fieldDef.Deprecated)).Id(transforms.ExportedFieldName(fieldDef.Name)).Add(fieldDef.Type.Code()).Tag(fieldTags)
 		}
-	})
+	}
+
+	// If there are docs, add them without the trailing line since the safety comment will alreaady add the trailing line
+	if objectDef.Docs != "" {
+		file.Comment(string(objectDef.Docs))
+	}
+	file.Add(safetyComment)
+	file.Type().Id(objectDef.Name).StructFunc(structFunc)
 
 	// If there are no collections, we can defer to the default json behavior
 	// Otherwise we need to override MarshalJSON and UnmarshalJSON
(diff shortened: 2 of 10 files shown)


conjure/objectwriter.go
```diff
@@ -28,10 +29,28 @@ const (
 	dataVarName     = "data"
 )
 
+// To avoid re-computing safety recursively, we keep the computed result in this global cache.
+var safetyCache = make(map[types.Type]spec.LogSafety)
```
nmiyake: Can you add this as a parameter to the relevant functions (for example, `getFieldSafety(fieldType types.Type, safetyCache map[types.Type]spec.LogSafety) spec.LogSafety`) and pass it in as a parameter?

Would prefer to avoid having global cache variables since it can make code hard to reason about (issues like concurrency guarantees, cache variable being shared by multiple tests if they have the same process, etc.)

conjure/objectwriter.go
```diff
@@ -28,10 +29,28 @@ const (
...
+
 func writeObjectType(file *jen.Group, objectDef *types.ObjectType) {
 	// Declare struct type with fields
 	containsCollection := false // If contains collection, we need JSON methods to initialize empty values.
-	file.Add(objectDef.Docs.CommentLine()).Type().Id(objectDef.Name).StructFunc(func(structDecl *jen.Group) {
+
+	// Compute the overall safety of the struct and add comment-based annotation
+	overallSafety := computeStructSafety(objectDef.Fields)
+	var safetyComment jen.Code = jen.Null()
```
nmiyake, line 52: nit, but I think that `jen.Null` and `nil` are handled in the same manner, so I think it might be a bit more canonical to just have `var safetyComment jen.Code`

conjure/objectwriter.go
```diff
@@ -28,10 +29,28 @@ const (
...
+	if !overallSafety.IsUnknown() {
+		switch overallSafety.Value() {
+		case spec.LogSafety_SAFE:
+			safetyComment = jen.Comment("safelogging:@Safe")
+		case spec.LogSafety_UNSAFE:
+			safetyComment = jen.Comment("safelogging:@Unsafe")
+		case spec.LogSafety_DO_NOT_LOG:
+			safetyComment = jen.Comment("safelogging:@DoNotLog")
+		}
```
nmiyake: Note that, since Go doesn't have sum types, if a new type is added here it will just be unhandled and `safetyCommnent` will be blank.

This could go either way, but given that this is within the context of a CLI/developer tool, I think it may make sense to have a `default` fallthrough case that panics

conjure/objectwriter.go
```diff
@@ -45,12 +64,32 @@ func writeObjectType(file *jen.Group, objectDef *types.ObjectType) {
...
+			if !fieldSafety.IsUnknown() {
+				switch fieldSafety.Value() {
+				case spec.LogSafety_SAFE:
+					fieldTags["safelogging"] = "@Safe"
+				case spec.LogSafety_UNSAFE:
+					fieldTags["safelogging"] = "@Unsafe"
+				case spec.LogSafety_DO_NOT_LOG:
+					fieldTags["safelogging"] = "@DoNotLog"
+				}
+			}
```
nmiyake: This logic is basically duplicated here and in the above block. I think it makes sense to decomp a function that does the safety -> string representation translation ("@Safe", "@Unsafe", "@DoNotLog", or blank/nil) and then logic here and in the other place will share representation and can remove redundancy of assignment logic in different cases

conjure/testdata/safety-tags-test.yml
```diff
@@ -0,0 +1,34 @@
...
+        alias: string
+        safety: unsafe
+      DoNotLogAlias:
+        alias: string
+        safety: do-not-log
+      NestedSafetyObject:
+        fields:
+          safeData: SafeAlias
+          unsafeData: UnsafeAlias
+          secretData: DoNotLogAlias
```
nmiyake: nit, but end file with newline

nmiyake (changes requested):

conjure/testdata/safety-tags-test.yml
sonnysasaka: Done.

conjure/objectwriter.go
sonnysasaka: Good idea, done.

conjure/objectwriter.go
sonnysasaka: Done.

conjure/objectwriter.go
sonnysasaka: I tried with `nil` but seems like jen crashes in that case, so we should use `jen.Null()` in this case, or handle `nil` specially. I prefer `jen.Null()` because there is no need for special handling it will just get ignored.

conjure/objectwriter.go
sonnysasaka: Done.

conjure/objectwriter.go
```diff
@@ -28,10 +29,33 @@ const (
...
+	}
+}
+
 func writeObjectType(file *jen.Group, objectDef *types.ObjectType) {
 	// Declare struct type with fields
 	containsCollection := false // If contains collection, we need JSON methods to initialize empty values.
-	file.Add(objectDef.Docs.CommentLine()).Type().Id(objectDef.Name).StructFunc(func(structDecl *jen.Group) {
+
+	// Compute the overall safety of the struct and add comment-based annotation
+	safetyCache := make(map[types.Type]spec.LogSafety)
```
nmiyake: Let's move cache up one more level (add as parameter to this function) so that it can be shared across all `writeObjectType` calls

conjure/testdata/safety-tags-test.yml
```diff
@@ -0,0 +1,34 @@
+types:
+  definitions:
+    default-package: test.safety
+    objects:
```
nmiyake, line 4: Can you add the following test cases as well:

* Object with list with item types that are safe/unsafe/do not log
* Object with set with item types that are safe/unsafe/do not log
* Object with map with key/value types that are safe/unsafe/do not log
* Alias type that doesn't declare type safety for itself but that aliases a different type with type safety declared properly inherits type safety

Also, what is the intended/desired behavior for an alias type that defines type safety but where referenced type also has type safety? Is the alias type allowed to make safety "more" permissive, or not?

conjure/objectwriter.go
```diff
@@ -93,3 +130,71 @@ func writeStructMarshalInitDecls(methodBody *jen.Group, fields []*types.Field, r
 		}
 	}
 }
+
+func getFieldSafety(fieldType types.Type, safetyCache map[types.Type]spec.LogSafety) spec.LogSafety {
+	// Check cache first
+	if cached, exists := safetyCache[fieldType]; exists {
+		return cached
+	}
```
nmiyake, line 141: Hmm, does this actually work?

The *type* of a field may hav an associated safety value.
However, a field can also declare a safety value itself (https://github.com/palantir/conjure/blob/master/docs/spec/conjure_definitions.md#fielddefinition), and this may vary/differ on a per-field basis (even with the same type), so it seems like you can't just store and retrieve based on type?

Hmm actually, based on the code at https://github.com/palantir/conjure-go/blob/6920a471840ccf71b96682f7276e11683393cfa4/conjure/types/types.go#L373, we may not have added per-field safety as a feature to `conjure-go` yet, in which case I guess this approach does work for now

conjure/objectwriter.go
```diff
@@ -93,3 +130,71 @@ func writeStructMarshalInitDecls(methodBody *jen.Group, fields []*types.Field, r
...
+
+	// Insert placeholder to detect cycles - use UNKNOWN as safe default
+	safetyCache[fieldType] = spec.New_LogSafety(spec.LogSafety_UNKNOWN)
```
nmiyake, line 144: It looks like this function (and its recursive calls) depend on `safetyCache` existing -- if that's the case, we may want to add a line at the beginning of the function that allocates a `safetyCache` map if the provided one is `nil` (so that the function still works/doesn't panic with a write to a `nil` map)

conjure/objectwriter.go
```diff
@@ -93,3 +130,71 @@ func writeStructMarshalInitDecls(methodBody *jen.Group, fields []*types.Field, r
...
+	// Default to SAFE, then find the most restrictive level
+	// Hierarchy: safe -> unannotated -> unsafe -> do-not-log
+	overallSafety := spec.New_LogSafety(spec.LogSafety_SAFE)
+
+	for _, fieldDef := range fields {
+		fieldSafety := getFieldSafety(fieldDef.Type, safetyCache)
+
+		if fieldSafety.IsUnknown() {
+			// Unannotated field "contaminates" the struct - it becomes unannotated
+			// unless we already found something more restrictive
```
nmiyake: I think "Field for which safety cannot be determined" is more accurate than "unannotated".

In the context of a struct, an "unannotated field" is a field that does not itself has an annotation. In the context of this comment, I think the meaning is that the safety for the field could not be determined (technically that would be because the type and all constituent types don't have annotations, but that's different from field not being annotated)

conjure/objectwriter.go
```diff
@@ -93,3 +130,71 @@ func writeStructMarshalInitDecls(methodBody *jen.Group, fields []*types.Field, r
...
+			if overallSafety.Value() == spec.LogSafety_SAFE {
+				overallSafety = spec.New_LogSafety(spec.LogSafety_UNKNOWN)
+			}
+		} else {
+			// Field has explicit safety annotation
```
nmiyake:
```suggestion
			// Specific safety value determined for field
```

Similar to above -- "the field is annotated" indicates to me that the field itself specifies/annotates a safety value, but in this context we're actually saying that the safety for the field was computed and was a non-unknown value

nmiyake (changes requested):

conjure/objectwriter.go
sonnysasaka: Done.

conjure/objectwriter.go
sonnysasaka: Done.

conjure/objectwriter.go
sonnysasaka: Done.

conjure/objectwriter.go
sonnysasaka: Now that we have safety from the field itself (not from the type), this is actually fine.

conjure/testdata/safety-tags-test.yml
sonnysasaka: Done. Also this reminds me that it was missing safelogging comments for alias which I just added.

As to for alias overriding the referenced type's safety, I don't find this is defined in the Conjure spec. So the current implementation simply will override regardless of the permissiveness.

conjure/objectwriter.go
sonnysasaka: Done.

conjure/errorwriter.go
```diff
@@ -59,7 +59,8 @@ func writeErrorType(file *jen.Group, def *types.ErrorDefinition) {
 func astErrorInternalStructType(file *jen.Group, def *types.ErrorDefinition) {
 	allArgs := append(append([]*types.Field{}, def.SafeArgs...), def.UnsafeArgs...)
 	// Use object generator to create a struct implementing JSON encoding for the error.
-	writeObjectType(file, &types.ObjectType{Name: transforms.Private(def.Name), Fields: allArgs})
+	safetyCache := make(map[types.Type]spec.LogSafety)
+	writeObjectType(file, &types.ObjectType{Name: transforms.Private(def.Name), Fields: allArgs}, safetyCache)
```
nmiyake, line 63: Fine to leave this as-is, but now that we added the logic to create a map if it's nil, you could just pass in `nil` for `safetyCache` here instead of instantiating

conjure/objectwriter.go
```diff
@@ -93,3 +129,85 @@ func writeStructMarshalInitDecls(methodBody *jen.Group, fields []*types.Field, r
...
+func getSafetyFromField(field *types.Field, safetyCache map[types.Type]spec.LogSafety) spec.LogSafety {
+	// First check if the field has an explicit safety annotation
+	if field.Safety != nil {
+		return *field.Safety
+	}
+	// Fall back to type-based safety
+	return getSafetyFromType(field.Type, safetyCache)
+}
+
+func computeStructSafety(fields []*types.Field, safetyCache map[types.Type]spec.LogSafety) spec.LogSafety {
```
nmiyake: Sorry, missed this in the first review.

Let's be more precise with the language/naming and parameter -- this function is computing the safety for a particular Conjure Object type, so I think the following naming for function and parameter is clearer:

```suggestion
func computeObjectSafety(object *types.ObjectType, typeSafetyCache map[types.Type]spec.LogSafety) spec.LogSafety {
```

`struct` is not a Conjure concept -- the `conjure-go` generator does generate `struct` code for Conjure objects, so for functions dealing with the generated struct code it makes sense to have `struct` as part of naming, but here it should be `object`

nmiyake (changes requested):
Thanks -- some feedback

conjure/objectwriter.go
sonnysasaka: Done.

conjure/errorwriter.go
sonnysasaka: Ack.

nmiyake (approved):

PR #719, opened 2025-07-18
feature: embed `extensions.conjure.json` in `embed.conjure.go` if `extensions` is present

aviradinsky:
This change modifies conjure-go to automatically embed an `extensions.conjure.json` file alongside generated Go code when the source Conjure IR contains [`extensions`](https://github.com/palantir/conjure/blob/master/docs/spec/intermediate_representation.md#extensions) metadata. This enables downstream tooling to access extension information that would otherwise be lost during code generation.

## Changes
**New file generation**: When `ConjureDefinition.Extensions` is non-empty, conjure-go now generates:
  - `extensions.conjure.json`: Contains the serialized extensions metadata from the Conjure IR
  - `embed.conjure.go`: Uses `Go 1.16+`'s `//go:embed` directive to embed the JSON file as a byte slice

## Technical Details
- Uses `safejson.MarshalIndent` for consistent, readable JSON formatting, in order to minimize git diffs and make the JSON more readable
- Refactored `OutputFile` structure to use a `render` function approach, enabling support for both jen-generated Go files and raw file content
- The `embed.conjure.go` file is generated using jen with an anonymous `embed` import and `//go:embed` directive

## ⚠️ Potential Linter Issues
The generated `embed.conjure.go` file includes a blank import `_ "embed"` which is required for the `//go:embed` directive. Some linters may flag this with the error `a blank import should be only in a main or test package, or have a comment justifying it`. This cannot be easily avoided **when using `jen`** since `jen` doesn't support adding comments to imports ([dave/jennifer#65](https://github.com/dave/jennifer/issues/65)). 

## Backward Compatibility
This change is fully backward compatible - existing generated code continues to work unchanged, and the new files are only generated when extensions are present in the source IR. The only risk is the linter issue mentioned above.

## Dependency Updates
This PR also includes dependency updates to resolve local build issues I encountered (possibly related to `github.com/pierrec/lz4 v2.0.5+incompatible`). All dependencies have been bumped to their latest versions. If preferred, I can split these dependency updates into a separate PR - just let me know!

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/719)


conjure/outputs.go
```diff
@@ -50,16 +50,20 @@ func (f *OutputFile) Write() error {
 }
 
 func (f *OutputFile) Render() ([]byte, error) {
-	f.file.HeaderComment("This file was generated by Conjure and should not be manually edited.")
+	bytes, err := f.render()
+	return bytes, errors.Wrapf(err, "failed to render bytes for file %s", f.absPath)
+}
 
+func renderJenFile(f *jen.File) ([]byte, error) {
+	f.HeaderComment("This file was generated by Conjure and should not be manually edited.")
```
nmiyake: I know this is preexisting, but the "official" generated file comment content is supposed to match the form `^// Code generated .* DO NOT EDIT\.$` ([link](https://github.com/golang/go/issues/13560#issuecomment-288457920)), so we should probably change to `"Code generated by conjure-go. DO NOT EDIT."`

Probably makes sense to do as a separate PR

conjure/outputs.go
```diff
@@ -50,16 +50,20 @@ func (f *OutputFile) Write() error {
 }
 
 func (f *OutputFile) Render() ([]byte, error) {
-	f.file.HeaderComment("This file was generated by Conjure and should not be manually edited.")
+	bytes, err := f.render()
+	return bytes, errors.Wrapf(err, "failed to render bytes for file %s", f.absPath)
```
nmiyake: nit, but slight preference for the following:

```suggestion
	bytes, err := f.render()
	if err != nil {
	   return nil, errors.Wrapf(err, "failed to render bytes for file %s", f.absPath)
	}
	return bytes, nil
```

It is true that, functionally, this is almost identical to what you have, since `errors.Wrapf` returns `nil` if `err` is `nil`, and this construction is more verbose.

However, for myself, over time I've found that making this explicit reduces cognitive overhead, and when debugging it also makes it easier to set break points/be able to tell immediately whether the "success" or "failure" logic branch is being executed.

conjure/conjure.go
```diff
@@ -124,6 +126,26 @@ func GenerateOutputFiles(conjureDefinition spec.ConjureDefinition, cfg OutputCon
 			}
 			files = append(files, newGoFile(filepath.Join(pkg.OutputDir, "servers.conjure.go"), serverFile))
 		}
+		if len(def.Extensions) > 0 {
+			extensions := "extensions.conjure.json"
+			{
```
nmiyake: Why are these written as separate clauses/blocks? Asking since we don't use this style very often.

conjure/conjure.go
```diff
@@ -124,6 +126,26 @@ func GenerateOutputFiles(conjureDefinition spec.ConjureDefinition, cfg OutputCon
...
+				extensionsFile := filepath.Join(pkg.OutputDir, extensions)
+				extensionsContent, err := safejson.MarshalIndent(def.Extensions, "", "\t")
+				if err != nil {
+					return nil, errors.Wrapf(err, "failed to marshall the conjure IR `extensions` field")
```
nmiyake:
```suggestion
					return nil, errors.Wrapf(err, "failed to marshal the conjure IR `extensions` field")
```

conjure/conjure.go
```diff
@@ -124,6 +126,26 @@ func GenerateOutputFiles(conjureDefinition spec.ConjureDefinition, cfg OutputCon
...
+				}
+
+				files = append(files, newRawFile(extensionsFile, extensionsContent))
+			}
+			{
+				embedFile := newJenFile(pkg, def, errorRegistryImportPath)
+				embedFile.Anon("embed")
+				embedFile.Comment(fmt.Sprintf("//go:embed %s", extensions))
+				embedFile.Var().Id("_").Index().Byte()
```
nmiyake: Let's factor this out into its own method for clarity/readability

nmiyake (changes requested):
Thanks! Generally looks good, but some minor feedback.

conjure/outputs.go
aviradinsky: 2aa3e01950ec195a6f1487e7077ce5427f4cb791

conjure/conjure.go
aviradinsky: 6c08ee395d1b71200a7484079bb04984202d054f

conjure/conjure.go
aviradinsky: The main reason I originally wrote these as separate blocks was to keep the variable scope as tight as possible. I find that this can help make the code a bit easier to follow, since variables only exist where they're actually needed, and it reduces the risk of accidentally reusing or shadowing them. It also lets me break the function into smaller, logical chunks, which I think can sometimes make complex code more digestible.

That said, it isn’t really the style we use elsewhere, so I reverted it in 3b3ef4cfabe8ec7391f7aa1b6504a4906c1378e3 to keep things consistent. If you prefer the block-based approach for clarity, I’m happy to switch back—just let me know! Otherwise feel free to `Resolve conversation`.

conjure/conjure.go
aviradinsky: 91ba24c5

conjure/outputs.go
aviradinsky: https://github.com/palantir/conjure-go/pull/720

conjure/conjure.go
```diff
@@ -124,6 +125,21 @@ func GenerateOutputFiles(conjureDefinition spec.ConjureDefinition, cfg OutputCon
 			}
 			files = append(files, newGoFile(filepath.Join(pkg.OutputDir, "servers.conjure.go"), serverFile))
 		}
+		if len(def.Extensions) > 0 {
+			extensions := "extensions.conjure.json"
```
nmiyake:
```suggestion
			const extensionsFileName = "extensions.conjure.json"
```

nit, but prefer the above (either inline as it currently is, or as a package-level variable) since it makes it more semantically clear what the value is and to make it clear that this is a constant rather than a mutable variable

conjure/conjure.go
```diff
@@ -124,6 +125,21 @@ func GenerateOutputFiles(conjureDefinition spec.ConjureDefinition, cfg OutputCon
...
+
+			extensionsFile := filepath.Join(pkg.OutputDir, extensions)
```
nmiyake: Since this isn't used elsewhere, probably makes sense to inline at call site in line 136 for consistency with how you handle this for `embed.conjure.go` at line 140

conjure/embedwriter.go
```diff
@@ -0,0 +1,27 @@
...
+	"fmt"
+
+	"github.com/dave/jennifer/jen"
+)
+
+func writeEmbedType(embed *jen.File, file string) {
+	embed.Anon("embed")
+	embed.Comment(fmt.Sprintf("//go:embed %s", file))
+	embed.Var().Id("_").Index().Byte()
+}
```
nmiyake:
```suggestion
// writeEmbedFileAsByteSlice adds code to the provided jen.File to embed the file at the specified path into a
// package-level with the specified variable name using the `//go:embed` directive. Also adds an import for the "embed"
// package.
func writeEmbedFileAsByteSlice(embed *jen.File, varName, filePath string) {
	embed.Anon("embed")
	embed.Comment(fmt.Sprintf("//go:embed %s", filePath))
	embed.Var().Id(varName).Index().Byte()
}
```

A few suggestions here:

* Change function name -- in other files, when we have `write*Type()`, that's typically because that function is writing code to define a Conjure type. In this instance, this function is writing code to embed a file using the `embed` directive, so function naming should reflect that.
* Technically the type for an `embed` statement could be `[]byte`, `string`, or `embed.Fs`. It's fine that we're not making that configurable, but would update function name and documentation to make this clear
* Similarly, our use case only requires an unnamed variable, but technically the variable could be named. Would either update function to allow variable name to be specified as a parameter, or update function name/documentation to reflect this behavior choice.

nmiyake (changes requested):
Thanks! A few more minor points of feedback.

conjure/conjure.go
aviradinsky: 1480daf1

conjure/embedwriter.go
aviradinsky: d215f3ee

conjure/conjure.go
aviradinsky: 6bcb564a

nmiyake (dismissed):

nmiyake (approved):

aviradinsky:
Released in `v6.68.0`

PR #725, opened 2025-08-04
✨ AcceptFuncs support for generic unions

glynternet:
This PR implements `AcceptFuncs` method generation for paremeterised/generic unions (`ThingWithT[T]`), bringing feature parity with non-parameterised/generic unions. The `AcceptFuncs` method provides an alternative to the visitor pattern that uses function parameters instead of interface implementations.

- **Added `AcceptFuncs` method generation** for generic unions that accepts function parameters returning `(T, error)`
- **Added helper methods** (`*NoopSuccess`, `ErrorOnUnknown`) for generic unions returning `(T, error)`
- **Added conditional generation** based on `cfg.GenerateFuncsVisitor` flag (matches non-generic behavior)
- **Comprehensive test coverage** including optional field handling and configuration scenarios

- **Return type**: `(T, error)` instead of just `error`

For a union with `string` and `int` fields:
```go
func (u *MyUnionWithT[T]) AcceptFuncs(
    strFunc func(string) (T, error),
    intFunc func(int) (T, error),
    unknownFunc func(string) (T, error)
) (T, error) {
    var result T
    switch u.typ {
    case "str":
        if u.str == nil {
            return result, fmt.Errorf("field \"str\" is required")
        }
        return strFunc(*u.str)
    // ... other cases
    }
}

func (u *MyUnionWithT[T]) StrNoopSuccess(string) (T, error) {
    var result T
    return result, nil
}

func (u *MyUnionWithT[T]) ErrorOnUnknown(typeName string) (T, error) {
    var result T
    return result, fmt.Errorf("invalid value in union type. Type name: %s", typeName)
}
```

- **Enabled**: When `GenerateFuncsVisitor: true` in conjure configuration
- **Disabled**: When `GenerateFuncsVisitor: false` (default behavior unchanged)

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/725)

### DIFF
--- a/changelog/@unreleased/pr-725.v2.yml
+++ b/changelog/@unreleased/pr-725.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: ✨ AcceptFuncs support for generic unions
+  links:
+  - https://github.com/palantir/conjure-go/pull/725
--- a/changelog/@unreleased/pr-725.v2.yml
+++ b/changelog/@unreleased/pr-725.v2.yml
@@ -1,5 +1,6 @@
 type: improvement
 improvement:
-  description: ✨ AcceptFuncs support for generic unions
+  description: ✨ AcceptFuncs support for generic unions, conditionally added with
+    the same accept-funcs configuration that already exists for non-generic unions.
   links:
   - https://github.com/palantir/conjure-go/pull/725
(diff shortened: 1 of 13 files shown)


nmiyake (approved):
Thanks, this looks great!

This looks good, but adding "do not merge" just so we can get another pair of eyes on this before merge.

bmoylan (approved):

PR #747, opened 2025-08-20
feature: filter conjure extensions to only `recommended-product-dependencies`

aviradinsky:
## Before this PR
A lot of noise was being created by `extension` keys that were not being consumed.

## After this PR
We now filter out all `extension` keys except for those that we expect to consume.

==COMMIT_MSG==
filter conjure extensions to only `recommended-product-dependencies`
==COMMIT_MSG==

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/747)

### DIFF
--- a/conjure-api/conjure-api-4.35.0.conjure.json
+++ b/conjure-api/conjure-api-4.35.0.conjure.json
@@ -1325,6 +1325,7 @@
   } ],
   "services" : [ ],
   "extensions" : {
+    "not-recommended-product-dependencies" : [ ],
     "recommended-product-dependencies" : [ ]
   }
 }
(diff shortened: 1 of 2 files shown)


conjure-api/conjure-api-4.35.0.conjure.json
```diff
@@ -1325,6 +1325,7 @@
   } ],
   "services" : [ ],
   "extensions" : {
+    "not-recommended-product-dependencies" : [ ],
```
aviradinsky, line 1328: This addition not showing up in https://github.com/palantir/conjure-go/blob/6d6176a207ee8560508433cf2cd6ee1896874cef/conjure-api/conjure/spec/extensions.conjure.json verifies that the change in this MR works.

bmoylan (approved):

PR #787, opened 2025-09-26
Revert "improvement: embed `extensions.conjure.json` on a per IR basi…

aviradinsky:
…s (#753)"

This reverts commit fb80c4511056cc8b48c1830e4c4829eb814464e7.

```
# Conflicts:
#	conjure/conjure.go
#	conjure/types/output_paths.go
```

### Problem
The previous implementation (#719) embedded `extensions.conjure.json` files on a per-IR basis, which caused build failures when multiple Conjure projects shared the same output directory and differing extensions in their IRs. This had the potential to affect ~36% of internal Conjure-using repositories, though the actual impact was likely less as conflicts only arise when repos have both conflicting output directories and differing extensions in their IRs.

### Root Cause
When multiple Conjure projects write to the same output directory, each project generates its own `extensions.conjure.json` file with potentially different content. During `./godelw conjure --verify`, the checksum validation fails because the file on disk gets overwritten by one project's extensions while regeneration produces extensions from another project.

### Follow-up
We are tracking work to disallow conflicting output directories in a future release. Once that enforcement is in place, we can revert back to per-IR extensions embedding for better isolation between projects.

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/787)


bmoylan (dismissed):

bmoylan (approved):

aviradinsky:
Released in `v6.77.0` 7ee80f3dea482cf035289fcdf5748f9f389c8774

aviradinsky:
`Release v6.88.0` https://github.com/palantir/godel-conjure-plugin/commit/f77b82d3b90dad0088c2db6e84cb3a4d77d3e3bf

PR #825, opened 2025-11-24
Update GenerateOutputFiles function

nmiyake:
## After this PR

==COMMIT_MSG==
Makes it such that the GenerateOutputFiles function no longer performs delete operations on disk. Also adds a WriteOutputFiles function and adds documentation to functions.
==COMMIT_MSG==

## Possible downsides?

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/825)

### DIFF
--- a/conjure/conjure.go
+++ b/conjure/conjure.go
@@ -15,8 +15,6 @@
 package conjure
 
 import (
-	"fmt"
-	"os"
 	"path"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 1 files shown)


conjure/conjure.go
```diff
@@ -67,27 +89,6 @@ func GenerateOutputFiles(conjureDefinition spec.ConjureDefinition, cfg OutputCon
...
-	for _, fileName := range [...]string{
-		"extensions.conjure.json",
-		"embed.conjure.go",
-	} {
-		filePath := filepath.Join(extensionsOutputDir, fileName)
-		if err := os.Remove(filePath); err != nil && !os.IsNotExist(err) {
-			return nil, fmt.Errorf("failed when attempted to remove file: %s: %w", filePath, err)
-		}
-	}
-
```
nmiyake, line 90: I considered factoring out this logic into a `RemoveDeprecatedFiles` exported function that could be called by clients that want to do this, but given the other updates I don't think this is actually needed, is it? For the newer plugin/config, "delete generated files" should delete files like this that still exist.

If we do think that it's important to still do this, I'll add an exported function for it (and probably call from `Generate` as well).

PR #826, opened 2025-11-24
fix: ensure consistent output when calling `Render()` multiple times

aviradinsky:
## Summary
Fixed an issue where calling `Render()` multiple times on the same `OutputFile` would produce different results. The generated header comment was being appended on each render, causing the output to grow with each call.

## Changes
- Moved the `HeaderComment` call from `renderJenFile()` to `newGoFile()`
- This ensures the header is added only once when the file is created, rather than on every render
- Added test case `TestOutputFileMultipleRenders` to verify idempotent rendering behavior

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/826)


nmiyake (approved):
Looks good, thanks!

PR #841, opened 2025-12-17
Feature: Allow generated code to use WGS/CGR V2 or V3 module versions

tabboud:
## Before this PR
Generated code only used WGS/CGR v2 module versions which blocked the migration of the new V3 modules. A major version is usually how new module versions are introduced, but given ongoing work in `godel-conjure-plugin` we want to avoid a major rev right now. Alternatively, we can expose a configuration point to allow controlling which module versions are used such that users can generate v3 module code if they migrate to use WGS/CGR v3 for the rest of the codebase

## After this PR

==COMMIT_MSG==
Allow generated code to use WGS/CGR V2 or V3 module versions
==COMMIT_MSG==

## Possible downsides?

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/841)

### DIFF
--- a/conjure/snip/imports.go
+++ b/conjure/snip/imports.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 
 	"github.com/dave/jennifer/jen"
+	werror "github.com/palantir/witchcraft-go-error"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


nmiyake (approved):

conjure/snip/imports.go
```diff
@@ -15,44 +15,85 @@
...
-	pal + "witchcraft-go-tracing/wzipkin":  "wzipkin",
-	wgs + "witchcraft/wresource":           "wresource",
-	wgs + "wrouter":                        "wrouter",
-	"gopkg.in/yaml.v3":                     "yaml",
-	"github.com/spf13/cobra":               "cobra",
-	"github.com/spf13/pflag":               "pflag",
+var (
+	// cgrModuleVersion controls which conjure-go-runtime module version to use in generated code.
+	// Defaults to 2. Set via SetCGRVersion before generation.
+	cgrModuleVersion = 2
```
nmiyake, line 31: Ultimately okay with this in the name of iteration speed, but I think it would be ideal to have a follow-up PR where we properly refactor the logic here such that these values are controlled and used per generation call rather than having them as package-level variables. Although this logic is currently only really used in CLIs in a single-threaded context, having package-level state makes it such that the code generation doesn't support concurrency and makes the operation generally more fragile.

PR #902, opened 2026-03-24
Break: Update module to v7

tabboud:
## After this PR

==COMMIT_MSG==
Update module to v7
==COMMIT_MSG==

## Possible downsides?

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/902)

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -17,7 +17,7 @@ package cmd
 import (
 	"strings"
 
-	"github.com/palantir/conjure-go/v6/conjure"
+	"github.com/palantir/conjure-go/v7/conjure"
 	"github.com/palantir/pkg/cobracli"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
--- a/conjure-api/generate.go
+++ b/conjure-api/generate.go
@@ -17,7 +17,7 @@
 package main
 
 import (
-	"github.com/palantir/conjure-go/v6/conjure"
+	"github.com/palantir/conjure-go/v7/conjure"
 )
 
 //go:generate go run $GOFILE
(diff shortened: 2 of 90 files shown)


nmiyake (commented):
Code change looks good.

Can you remind me where the breaking change will come into play? If `godel-conjure-plugin` updates to use this version, then does the code it generates start to break again?

We likely discussed this, but just wanted a refresher on what rollout looks like here.

tabboud:
The plan is as follows:
- Release `conjure-go` v7.0.0
- Update `godel-conjure-plugin` to use `conjure-go` v7
- Release `godel-conjure-plugin` v7.0.0 (major version gate for new generated code)

Automation should not update the major version of the godel plugin (I will double check), so when we move WGS to g.p.b, the excavator that will migrate projects will also update the `godel-conjure-plugin` to use the v7 version. When WGS is moved we are also going to drop the deprecated `wrouter` and `wresource` so the generated code will line up with the usages of `witchcraft-go-router` in WGS.

nmiyake:
OK thanks -- I would double-check on the assumption that automation won't bump the plugin version across major versions. Although that's probably something that we *should* do, I don't think that we do right now (pretty sure plugin/asset update logic in automation just checks latest available version and upgrades). However, I think that the `godel-conjure-plugin` update is broken out into its own check -- if that's the case then it shouldn't be too bad, since even if the PR/check is blocked for generating something that doesn't compile, it won't block other plugin/asset updates (that's the main concern)

nmiyake:
Either way, approving this piece shouldn't impact that, so we can proceed.

nmiyake (approved):
Code change looks good.

Can you remind me where the breaking change will come into play? If `godel-conjure-plugin` updates to use this version, then does the code it generates start to break again?

We likely discussed this, but just wanted a refresher on what rollout looks like here.

PR #931, opened 2026-05-13
Add --strict-unmarshal flag to reject unknown JSON fields

ashrayjain:
## Before this PR
We did not have the ability to generate code with strict unmarshaling

## After this PR

==COMMIT_MSG==
Objects for which conjure generates custom JSON unmarshallers support optional rejection of unknown fields
==COMMIT_MSG==

## Possible downsides?
This doesn't automatically work for objects that we dont generate custom unmarshallers for and for those, the consumer needs to set the appropriate flags on the JSON decoder they use. With new golang Json encoder library work, there will be a way to get decoder options piped through to custom unmarshaller code, at which point we can switch conjure-go to use that and deprecate/remove the CLI flag that is being added in this PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/931)

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -24,17 +24,19 @@ import (
 )
 
 const (
-	outputDirFlagName    = "output"
-	serverFlagName       = "server"
-	funcsVisitorFlagName = "funcs-visitor"
+	outputDirFlagName        = "output"
+	serverFlagName           = "server"
+	funcsVisitorFlagName     = "funcs-visitor"
+	strictUnmarshalFlagName  = "strict-unmarshal"
 )
 
 var (
-	version             = "unspecified"
-	debug               bool
-	outputDirFlagVar    string
-	serverFlagVar       bool
-	funcsVisitorFlagVar bool
+	version                = "unspecified"
+	debug                  bool
+	outputDirFlagVar       string
+	serverFlagVar          bool
+	funcsVisitorFlagVar    bool
+	strictUnmarshalFlagVar bool
 )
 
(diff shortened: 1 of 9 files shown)


cmd/root.go
```diff
@@ -24,17 +24,19 @@ import (
 )
 
 const (
-	outputDirFlagName    = "output"
-	serverFlagName       = "server"
-	funcsVisitorFlagName = "funcs-visitor"
+	outputDirFlagName       = "output"
+	serverFlagName          = "server"
+	funcsVisitorFlagName    = "funcs-visitor"
+	strictUnmarshalFlagName = "strict-unmarshal"
```
iamdanfox, line 30: 👎 I think giving people a single flag to set this for ALL generated objects is too much of a foot gun.

Suppose someone uses this flag in a project to generate CLIENT BINDINGS from some conjure IR to let them interact with an upstream server. Then, the upstream server decides to add one more field to their response objects. 

The client (with strict unmarshal) wouldn't know about the new response field and would break!!

This is bad, and would violate the foundational backcompat expectations that the conjure wire format requires people to adhere to: https://github.com/palantir/conjure/blob/master/docs/spec/wire.md#41-forward-compatible-clients

iamdanfox: Maybe one way we could safely to do this is if you analyze the Conjure IR and figure out which objects are ONLY used in requests... then you could say for these request-only objects, it's fine to use strict unmarshalling.

PR #971, opened 2026-06-15
Feature: Add ErrorParameterFormatJSON option

tabboud:
## Before this PR
There was no ability to set the JSON error parameter format for generated clients.

## After this PR

==COMMIT_MSG==
Add ErrorParameterFormatJSON option
==COMMIT_MSG==

When `OutputConfiguration.ErrorParameterFormatJSON` is set to true, all requests will set the `Accept-Conjure-Error-Parameter-Format: JSON` header so that servers serialize error parameters as JSON. This is opt-in only since there could be behavior that users depend on with the legacy error parameter format. Servers that do not support this treat this header as a noop.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/dave/jennifer v1.7.1
 	github.com/julienschmidt/httprouter v1.3.0
 	github.com/nmiyake/pkg/dirs v1.1.0
-	github.com/palantir/conjure-go-runtime/v3 v3.15.1-0.20260615194028-9bc18ed2425e
+	github.com/palantir/conjure-go-runtime/v3 v3.18.0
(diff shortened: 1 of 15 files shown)


go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/dave/jennifer v1.7.1
 	github.com/julienschmidt/httprouter v1.3.0
 	github.com/nmiyake/pkg/dirs v1.1.0
-	github.com/palantir/conjure-go-runtime/v3 v3.15.0
+	github.com/palantir/conjure-go-runtime/v3 v3.15.1-0.20260615194028-9bc18ed2425e
```
tabboud: Depends on https://github.com/palantir/conjure-go-runtime/pull/935 and will be switched once that merges

nmiyake (approved):

PR #1013, opened 2026-08-05
fix: Format generated files without goimports' import resolution

bmoylan:
Each ptimports call previously built a fresh resolver that shelled out to "go env" and read the entire module-cache index before concluding the file was already complete. Regenerating an 11-project, 617-file tree drops from 2m25s to 9s. Generated output is byte-identical.

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-go/1013)

### DIFF
--- a/conjure/conjure.go
+++ b/conjure/conjure.go
@@ -19,8 +19,11 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"regexp"
 	"sort"
+	"strconv"
+	"strings"
+	"unicode"
+	"unicode/utf8"
 
 	"github.com/dave/jennifer/jen"
 	"github.com/palantir/conjure-go/v7/conjure-api/conjure/spec"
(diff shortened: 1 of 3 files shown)


conjure/conjure.go
```diff
@@ -227,6 +226,49 @@ func newRawFile(filePath string, bytes []byte) *OutputFile {
...
+func importPathToAssumedName(importPath string) string {
+	base := path.Base(importPath)
+	if strings.HasPrefix(base, "v") {
+		if _, err := strconv.Atoi(base[1:]); err == nil {
+			if dir := path.Dir(importPath); dir != "." {
+				base = path.Base(dir)
+			}
+		}
+	}
+	base = strings.TrimPrefix(base, "go-")
```
bmoylan, line 262: I don't really like this but it matches ptimports

tabboud (approved):
👍

